PR #1, opened 2021-05-14
Try using GITHUB_TOKEN for auth

jonathangreen:
**What's this do?**

Using GITHUB_TOKEN seems too easy and it may just be scoped to the particular repo the request is coming from, but I thought it was worth a try.


jonathangreen:
Doesn't seem to have correct access :(

PR #6, opened 2021-05-20
Enable account-specificc documentation viewer

talesgurjao:
**What's this do?**
Allow users to view library documents from account screen 

https://www.notion.so/lyrasis/Allow-users-to-view-library-documents-from-account-screen-779d77a7d24d4794a42b833101cb308e

Try to add a new library (New York Public library) and check if you can open EULA, Privacy Policy and Content Licenses docs

n/a

n/a

@talesgurjao ran the Palace app.


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountDetailViewModel.kt
```diff
@@ -69,14 +71,20 @@ class AccountDetailViewModel(
...
 
-  val eula: EULAType? =
-    this.documents.eula
-
   val account =
     this.profilesController
       .profileCurrent()
       .account(this.accountId)
 
+  val eula: SimpleDocument? =
```
ray-lee, line 79: Is there any reason this is declared as `SimpleDocument` instead of `DocumentType`?

ray-lee (commented):
Just the one question above.

PR #12, opened 2021-06-29
Add the ability to specify a custom library registry

io7m:
**What's this do?**
This adds a debugging interface and a bit of UI in the debug settings
menu that allows for setting a custom library registry server. The
setting does not persist after an app restart.

Affects: https://www.notion.so/lyrasis/Add-ability-in-Palace-Android-app-to-configure-library-registry-libraries-URL-7f88c7e8d0c84efe9b6fcc9906e4605a

Try the debug menu!

None.

N/A

@io7m Tested setting various URIs, including some that aren't valid

device-2021-06-29-110051


io7m:
Note: When you set this URI, it'll clear the cache. On the accounts list screen, you'll need to select Reload from the toolbar to fetch new account definitions from the server.

ray-lee (approved):
Looks good to me!

PR #14, opened 2021-07-02
Fix SAML download endless loop.

ray-lee:
**What's this do?**

This adds a SAML download context object to the borrowing context, to track information about the current SAML download. This is used by the SAML download subtask. When that subtask runs, it may find HTML content, which means that it should exit so that a login can happen in a webview. Once login is complete, the download subtask will run again, this time with a SAML download context that indicates that authentication is complete, and contains a different URL to use to retrieve the download. The download can then proceed with the new URL instead of the original.

Fixes an endless loop that can occur when doing SAML downloads: https://www.notion.so/lyrasis/Borrowing-from-SAML-testbed-causes-endless-loop-56303fd7b2044498981b16d6cad5947a

Download a book from the "SAML" library. A form from the IdP should be displayed, to approve federating the authentication. When submitted, the download should succeed (instead of looping endlessly).

n/a

n/a

@ray-lee ran the Palace app.

### DIFF
--- a/simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/internal/BorrowSubtaskDirectory.kt
+++ b/simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/internal/BorrowSubtaskDirectory.kt
@@ -14,9 +14,10 @@ class BorrowSubtaskDirectory : BorrowSubtaskDirectoryType {
       BorrowAxisNow,
       BorrowBearerToken,
       BorrowCopy,
+      // BorrowSAMLDownload must precede BorrowDirectDownload in precedence.
+      BorrowSAMLDownload,
       BorrowDirectDownload,
       BorrowLCP,
-      BorrowLoanCreate,
-      BorrowSAMLDownload
+      BorrowLoanCreate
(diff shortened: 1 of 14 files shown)


simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/internal/BorrowSubtaskDirectory.kt
```diff
@@ -14,9 +14,10 @@ class BorrowSubtaskDirectory : BorrowSubtaskDirectoryType {
       BorrowAxisNow,
       BorrowBearerToken,
       BorrowCopy,
+      // BorrowSAMLDownload must precede BorrowDirectDownload in precedence.
+      BorrowSAMLDownload,
```
io7m, line 18: This is a good point. The semantics with regards to priority in subtask selection is undefined right now. Having it just be in the order they were passed in is not great... Something we need to fix.

io7m (approved):
This looks good to me! SAML makes me kind of nervous, so I can't speak to the validity of the download steps. :smile: 

The actual implementation looks good, though!

PR #16, opened 2021-07-07
Persist custom library registry to shared preferences

io7m:
**What's this do?**
This change makes the account registry debugging properties persistent
by using shared preferences. Typically we would _not_ use shared
preferences in the app, because they aren't profile-aware. In this
case, though, the debugging properties are truly app-global anyway.

Affects: https://www.notion.so/lyrasis/Android-build-that-can-persistently-uses-QA-library-registry-9874931996be425e8bbc378bcbbf3a1e

Check that the registry setting persists after an app restart.
Check that clearing the registry value results in a blank value after an app restart.

None.

N/A

@io7m Tested setting and unsetting, and restarting the app several times.


ray-lee (approved):
Looks good to me!

PR #19, opened 2021-07-15
Hide the default library from users.

ray-lee:
**What's this do?**

This hides the default library from being seen by users.

1. Configure the default library to be a “fake” library (one whose uuid isn’t the same as any library in the registry). This makes every library in the registry able to be selected.
1. When a library is added, automatically delete the default library from the profile. Since the user must select a library when the app starts (and before they can do anything else), this means they never see the default library.
1. Filter the default library out of the library selection list, so it can’t be added by the user in the future.

Fixes: https://www.notion.so/lyrasis/DPLA-library-doesn-t-appear-in-library-selection-list-31741e673f594e74b7fc54da0dc78e00

Also, users not understanding, or unintentionally being in, the default library has been causing support headaches.

On a fresh install of the app, tap the "Find Your Library" button. The "Digital Public Library of America" library should be selectable. After adding a library, there should only be one library in the list. Adding more libraries, and removing them, should continue to work normally.

n/a

n/a

@ray-lee ran the Palace app.

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <c:changelog project="org.thepalaceproject.palace" xmlns:c="urn:com.io7m.changelog:4.0">
   <c:releases>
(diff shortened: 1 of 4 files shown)


io7m:
This looks OK!

io7m (approved):

PR #21, opened 2021-07-17
Do not fail if position parsing fails

io7m:
**What's this do?**
This upgrades to a version of the audio books API that will not fail
to parse positions that contain explicitly `null` titles. It also logs
an error and returns nothing instead of loudly raising an exception
if the saved position turns out to be malformed.

Affects: https://www.notion.so/lyrasis/Error-opening-app-after-exiting-from-audiobook-player-5694f5b2d40c4baea0d97782a3141cef

Try the behaviour in the ticket, and make sure the app no longer goes insane.

We shouldn't release a version of the app pointing at a snapshot version of the audiobook API. A new release should be tagged there if all goes well.

N/A

@io7m's app refused to break

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,6 +120,7 @@ subprojects { project ->
   repositories {
     mavenLocal()
     maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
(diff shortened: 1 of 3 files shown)


ray-lee (approved):
Looks good, thanks!

PR #23, opened 2021-07-21
Gracefully handle expired certificates

io7m:
**What's this do?**
This updates the boot code and DRM service to check for and reject
expired certificates. It adds a new boot event type that allows for
displaying a dialog if something important but not fatal occurs on
boot, and then publishes an event if the DRM fails.

Affects: https://www.notion.so/lyrasis/Android-Properly-handle-expired-Adobe-certificate-427ab218209d40d3856246cb220fdbca
Affects: https://www.notion.so/lyrasis/App-crashes-on-login-when-Adobe-cert-is-expired-e71eda890b034802ae6ceab83d363d1a

If everything goes well, you should never see the effects of this code. If you want to see what happens, try using the currently-expired certificate.

None.

N/A

@io7m Used the expired certificate today.

### DIFF
--- a/simplified-adobe-extensions/src/main/java/org/nypl/simplified/adobe/extensions/AdobeDRMServices.java
+++ b/simplified-adobe-extensions/src/main/java/org/nypl/simplified/adobe/extensions/AdobeDRMServices.java
@@ -18,6 +18,7 @@ import com.io7m.jnull.NullCheck;
 import com.io7m.junreachable.UnimplementedCodeException;
 import com.io7m.junreachable.UnreachableCodeException;
 
+import org.joda.time.Instant;
 import org.nypl.drm.core.AdobeAdeptConnectorFactory;
 import org.nypl.drm.core.AdobeAdeptConnectorFactoryType;
 import org.nypl.drm.core.AdobeAdeptConnectorParameters;
@@ -89,8 +90,7 @@ public final class AdobeDRMServices {
    * certificate across differently branded versions of the same application
    * (with different package IDs) during development.
    *
-   * @param context          Application context
-
+   * @param context Application context
    * @return A DRM content filter
    * @throws DRMException If DRM is unavailable or cannot be initialized.
    */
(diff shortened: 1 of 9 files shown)


io7m:
I left this code using the old certificate to give you a chance to see what it does with the expired one. The SHA-256 will need to be updated to match that of the new certificate.

ray-lee (commented):
When I run this normally, the dialog doesn't appear. But if I run the debugger and put a breakpoint on `createAdobeExecutor` (line 1069), it does. Is there a race condition somewhere?

io7m:
Hrm! That's weird... If you add a logging statement in the event handler at https://github.com/ThePalaceProject/android-core/pull/23/files#diff-b036a508554756df415b9cf1d62bbc460dd59661f38e1402116c65f9910302f5R91, is the event definitely being handled? I'm not sure how/why it wouldn't be, but this is Android, after all.

io7m:
Wondering if it might be more reliable to use `requireActivity()` than `requireContext()` there. :thinking:

ray-lee:
I put in a log statement, and it looks like the event is not being handled.
It's getting here: https://github.com/ThePalaceProject/android-core/pull/23/files#diff-95986fb20b17a413886d1e9b6478bc321ceaadc8c96ba26555187e4dce0d1156R1072
But not here: https://github.com/ThePalaceProject/android-core/pull/23/files#diff-b036a508554756df415b9cf1d62bbc460dd59661f38e1402116c65f9910302f5R92

io7m:
Could you confirm that it still _doesn't_ work with this latest commit? If this doesn't work, then I think it is just a basic race condition as you suggested. I think the UI subscribes to the boot event stream after the event has already passed. Only the latest event is buffered, so it's possible to miss them.

I'll have to move the dialog elsewhere.

ray-lee:
Yup, it still doesn't work with that commit.

io7m:
Ok, this one forgoes using a boot event and instead just calls a function directly from the `MainFragment` `onStart` method. This does mean you won't see a dialog until after the app has started up, but I don't think that matters.

ray-lee (approved):
This looks good! The dialog does appear for me now.

PR #25, opened 2021-08-04
Start introducing common icons

io7m:
**What's this do?**
This adjusts the icons used in the tabs views to match iOS. The
profile tab has no corresponding icon and has been removed.

Affects: https://www.notion.so/lyrasis/Update-Android-UI-to-match-iOS-and-implement-new-icon-and-Brand-design-f504c6b7f5fe40c6a66c40a6e1c3438b

Check that the UI looks closer to the PDF. :smile: 

None.

N/A

@io7m did.

device-2021-08-04-152339


ray-lee (approved):
Looks good to me!

PR #26, opened 2021-08-27
Remove constraint on font scale when building reader preferences.

ray-lee:
**What's this do?**

Removes the constraint on the font scale when building reader preferences, which limits the value to be between 50 and 200.

This fixes a bug where the user can select a font scale greater than 200 in Readium, but it is limited to a max of 200 when the Readium settings are converted to user preferences. Since the user preferences are what get saved and restored when the user closes and re-opens the app, this causes the scale to be restored to 200, instead of the user's actual selection.

When the preferences are converted back to Readium settings, the font scale is already being constrained to the min and max values allowed by the Readium theme: https://github.com/ThePalaceProject/android-core/blob/04e0b9c26182dcd9f380c82aedb155d68d355f6f/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Themes.kt#L71-L75

So there's no need to have this additional constraint that isn't aware of what Readium allows.

Addresses part of this Notion ticket: https://www.notion.so/lyrasis/Font-size-button-Android-f49da8f869b84d1c80755c6c0e8c4937

Read an epub book. In the reader settings, select a large font size (at least 10 steps up from the default). Exit and restart the app. The font size should appear to be the same as it was before exiting.

n/a

n/a

@ray-lee ran the Palace app.

### DIFF
--- a/simplified-reader-api/src/main/java/org/nypl/simplified/reader/api/ReaderPreferences.java
+++ b/simplified-reader-api/src/main/java/org/nypl/simplified/reader/api/ReaderPreferences.java
@@ -101,7 +101,6 @@ public abstract class ReaderPreferences {
      */
 
     public final ReaderPreferences build() {
-      this.setFontScale(Math.min(200.0, Math.max(50.0, fontScale())));
       this.setBrightness(Math.min(1.0, Math.max(0.0, brightness())));
       return autoBuild();
     }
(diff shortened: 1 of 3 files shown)


io7m (approved):
Seems fine to me!

What, people don't like fonts at 800%? :laughing:

PR #28, opened 2021-09-01
Use the new palaceproject.io registry

io7m:
**What's this do?**
This sets the Palace app library registry URI to the new registry.

The old registry is being purged!

Check that there are accounts listed in the "Create An Account" section of the app. Right now, there are only non-production libraries in the registry, so you'll need to toggle the switch in the debug menu to enable testing/QA libraries.

None.

N/A

@io7m 

device-2021-09-01-182121


jonathangreen (approved):
Looks good to me! 🤘🎸

ray-lee (approved):
👍

PR #32, opened 2021-09-17
Automatically switch to catalog

io7m:
**What's this do?**
This adds some code to the main fragment listener delegate that
switches back to the catalog automatically when an account is created.

Affects: https://www.notion.so/lyrasis/Switch-library-catalog-when-adding-new-library-cc488ba037fa4704960b08cbd03262cc

1. Open the "Add a new library" screen
2. Select any library
3. Note that you're left looking at the Catalog, with the new library selected

None.

N/A

@io7m Tried it with a few libraries.


ray-lee (approved):
LGTM!

PR #34, opened 2021-09-22
Adjust toolbar handling

io7m:
**What's this do?**
This adjusts how the toolbar is handled for book detail fragments and for other
screens. It fixes two main problems:

  * The MainFragment's handling of the toolbar could override the handling
    implemented by other fragments when the activity is resumed.
  * The BookDetailFragment's handling of the toolbar was outright insufficient,
    and didn't implement the proper back button behaviour.

Affects: https://www.notion.so/lyrasis/Toolbar-changes-when-going-back-from-epub-9db3750264324816987e4df1de8baebd
Affects: https://www.notion.so/lyrasis/Back-button-does-not-work-when-trying-to-go-back-from-Book-Details-View-fb88a9c983c64c09818dd2d88dc921b7

Try the steps given in both tickets.

None.

N/A

@io7m tried the steps in the tickets.

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainFragment.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainFragment.kt
@@ -154,14 +154,6 @@ class MainFragment : Fragment(R.layout.main_tabbed_host) {
       .let { subscriptions.add(it) }
 
     /*
-     * Show the Toolbar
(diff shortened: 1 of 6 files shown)


nunommts (approved):
Great work here!

ray-lee (approved):
Looks good to me.

PR #36, opened 2021-09-23
Sync with upstream

ray-lee:
**What's this do?**

This merges in changes from the upstream ([NYPL-Simplified/Simplified-Android-Core](https://github.com/NYPL-Simplified/Simplified-Android-Core)) repo.

This gives us features and bug fixes that were implemented upstream, including:
- Update to latest Readium 2 libraries, so features from this update are available: https://github.com/ThePalaceProject/android-r2/pull/3
- Fixing more lifecycle issues
- Correctly removing expired loans

Use the app, and verify that features and bug fixes implemented for Palace still work, including:
- Documentation viewers in settings and on accounts
- Updated icons and color scheme
- Ability to set custom registry server
- Hiding the default library/forcing users to select a library on first run
- Automatically opening the catalog when a library is added

n/a

n/a

@ray-lee tested that the above Palace features still work.


ray-lee:
@io7m I don't expect you to go through every line of this unless you really want to, but if there's anything you want to check to be sure it wasn't mangled in the merge, please do.

@nunommts Mostly an FYI. This might produce conflicts in work you have in progress, since a lot of files are touched.

nunommts:
@ray-lee Thanks for the heads up! I usually do that so I can have the branches up-to-date but I'll check it twice this time :D

PR #37, opened 2021-09-23
Toolbar adjustments (again)

io7m:
**What's this do?**
This is a continuation of the previous [Adjust Toolbar Handling](https://github.com/ThePalaceProject/android-core/pull/34) PR but properly extended to all screens in the app.

The toolbar button in the Settings screen is configured to do nothing when it isn't a back arrow. This behaviour is easy to change if needed.

See the Notion tickets on the original PR.

Make sure the toolbar does something sensible on every screen in the app.

None.

N/A

@io7m tested every screen.


ray-lee (approved):
Looks good, thanks!

PR #41, opened 2021-09-30
Add support to featured libraries

nunommts:
**What's this do?**
This PR adds a list of the supported libraries ids to the _AccountListRegistryViewModel_ which will be responsible for ordering them by type (featured or not) and alphabetically in the case of being featured. If the list of available AccountProviders has at least one of the featured libraries, we'll add a _null_ item between the featured and non-featured so it can be distinguished in the list of libraries to add.
The PR also changes the _FilterableAccountListAdapter_ by making it handle a null item which will be the empty space in the final list. That adapter now distinguishes between a regular item type and a gap type and inflates the corresponding view.

These changes were done in order to replicate the iOS behavior of displaying the featured and regular libraries separated by an empty space when the user is adding a library. This request can be found [here](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=32b300218b974ecdb2054533a8b4a6d2&p=7d5060cf7ff6497183327c9a67c131d7).

_Logged in user_
Open the app;
Click on the top left icon;
Click on the "Add library" option of the dialog that shows up
See the featured libraries (Open Text Library and Palace Bookshelf), if you don't have them selected, at the top of the list, followed by an empty space and the following available libraries ([See screenshot](https://user-images.githubusercontent.com/79104027/135443332-2c62075d-1fa6-423e-aaad-9480eec5bebe.png))

_Non authenticated user_
Open the app;
Click on "Find Your Library" button
See the featured libraries (Open Text Library and Palace Bookshelf) at the top of the list, followed by an empty space and the following available libraries ([See screenshot](https://user-images.githubusercontent.com/79104027/135443332-2c62075d-1fa6-423e-aaad-9480eec5bebe.png))

No

Yes

No

Tested by @nunommts


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
```diff
@@ -189,7 +194,29 @@ class AccountListRegistryViewModel(private val locationManager: LocationManager)
         .filter { it.id != accountRegistry.defaultProvider.id }
         .toMutableList()
       updatedProvidersList.removeAll(usedAccountProviders)
-      accountProvidersList.onNext(updatedProvidersList)
+
+      val featuredLibrariesList = updatedProvidersList.filter {
+        featuredLibrariesList.contains(it.id.toString())
```
ray-lee, line 199: Shadowing `featuredLibrariesList` like this is confusing. I would at least put `this.` on the inner usage to make it clearer, but the best option is to use a different name for the local variable. (This may become moot if you move the definition of the featured libraries out of this file.)

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
```diff
@@ -48,6 +48,11 @@ class AccountListRegistryViewModel(private val locationManager: LocationManager)
   private val logger =
     LoggerFactory.getLogger(AccountListRegistryViewModel::class.java)
 
+  private val featuredLibrariesList = arrayOf(
```
ray-lee, line 51: Instead of having this here, I think this list should be defined in [PalaceBuildConfigurationService](https://github.com/ThePalaceProject/android-core/blob/main/simplified-app-palace/src/main/java/org/thepalaceproject/palace/PalaceBuildConfigurationService.kt), where we have other "configuration" settings for the app. It will be easier to find if we want to change it.

README-CHANGES.xml
```diff
@@ -89,6 +89,7 @@
           </c:tickets>
         </c:change>
         <c:change date="2021-09-28T00:00:00+00:00" summary="Fix long chapter titles overlapping elapsed/total time displays in audiobook player"/>
+        <c:change date="2021-09-30T11:37:51+00:00" summary="Added support to featured libraries"/>
```
ray-lee, line 92: Should be "Added support _for_ featured libraries". "Support to" sounds a bit weird here.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryFragment.kt
```diff
@@ -198,12 +196,14 @@ class AccountListRegistryFragment : Fragment(R.layout.account_list_registry) {
           }
           newText.equals("NYPL", ignoreCase = true) -> {
             this@AccountListRegistryFragment.accountListAdapter.filterList { account ->
-              account.title.contains("New York Public Library", ignoreCase = true)
+              account != null &&
+                account.title.contains("New York Public Library", ignoreCase = true)
```
ray-lee, line 200: Ha, I forgot this was in here. It's fine to keep it, but for Palace, it will be more of an Easter egg than anything actually useful.

ray-lee (changes requested):
This mostly looks fine. Just a few suggestions.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
nunommts: You're right. I didn't even notice the names were the same and that's why I kept it like this. I'll fix this

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
nunommts: I don't know how I've missed the Notion task description suggesting that file to be the one with this list. I'll move it.

README-CHANGES.xml
nunommts: Non-native english mistake, sorry :D

ray-lee (approved):
Looks good, thanks!

PR #42, opened 2021-10-01
Enable back button in reader toolbar.

ray-lee:
**What's this do?**

This enables the newly added back button in the reader toolbar. The latest build of android-r2 adds a back button to the reader toolbar, which fires the `SR2ReaderViewNavigationClose` event. Currently, the `SR2ReaderViewNavigationClose` event handler assumes that the TOC is open, and all it does is attempt to close the TOC. This adjusts the handler to actually return to the previous screen when the TOC is not open.

This makes the toolbar behave more like it does on iOS. It also makes it easier for users to exit a book when the OS-provided back button is not enabled. Notion: https://www.notion.so/lyrasis/Add-back-button-to-reader-toolbar-0f25bf9b71fc4eae99b042a521d34c7e

Read an epub book. If the toolbar is not visible, tap in the middle of the page to open it. On the left side of the toolbar, there should be a back arrow. Tapping on the back arrow should exit the book, and return to the screen from which the book was opened (the bookshelf or book detail screen). When reading a book, opening and closing the TOC should continue to work as normal.

https://github.com/ThePalaceProject/android-r2/pull/4 should be merged first. This PR won't do anything if that isn't merged first, but it won't do any harm either.

Yes. The changelog is updated with this change, as well as the other android-r2 change (reordering the toolbar menu items).

n/a

@ray-lee ran the Palace app.

### DIFF
--- a/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Activity.kt
+++ b/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Activity.kt
@@ -302,7 +302,7 @@ class Reader2Activity : AppCompatActivity(R.layout.reader2) {
 
     return when (event) {
       SR2ReaderViewNavigationClose ->
-        this.tocClose()
+        this.onBackPressed()
       SR2ReaderViewNavigationOpenTOC ->
         this.tocOpen()
       is SR2ControllerBecameAvailable ->
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -50,7 +50,7 @@
         <c:change date="2021-09-10T00:00:00+00:00" summary="Introduced a new color scheme for the application."/>
       </c:changes>
     </c:release>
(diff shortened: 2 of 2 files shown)


simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Activity.kt
```diff
@@ -302,7 +302,7 @@ class Reader2Activity : AppCompatActivity(R.layout.reader2) {
 
     return when (event) {
       SR2ReaderViewNavigationClose ->
-        this.tocClose()
+        this.onBackPressed()
```
ray-lee, line 305: This is a one line change. The existing `onBackPressed` method calls `this.tocClose()` if the TOC is open, or the superclass `onBackPressed` otherwise. The latter will pop back the stack.

https://github.com/ThePalaceProject/android-core/blob/9b839bd0460060718c29522bd4810cdec2cd84e3/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Activity.kt#L344-L350

nunommts (approved):

PR #43, opened 2021-10-01
Fix cropped button text on smaller devices

nunommts:
**What's this do?**
This PR changes how the text size value is set to both textviews on the button with the missing loan duration and how the clock image layout parameters are initialized. Instead of using direct values, we use the converter in order to get the corresponding value in _sp_ (for the text size) and in _dp_ (for the image).

[A bug](https://www.notion.so/lyrasis/Time-left-on-loan-being-cutoff-on-Android-6bc0f1a88ff64e969c58ae651b7c61a2) was reported saying the missing loan duration text was being cropped in some devices with smaller screens.

Open the "Books" tab;
On the catalog, if you have books with loan durations (for instance, in Lyrasis library, **Companions** from _Christina Hesselholdt_ and **In the Neighborhood of True** from _Susan Kaplan Carlton_) you can see the button "Read" / "Download" / "Listen" is displaying an image with a clock and the [missing loan duration without any cutoff text](https://user-images.githubusercontent.com/79104027/135688991-3f65f0dc-fb59-4fed-ab8a-bdb19f8c1174.png).

n/a

Tested by @nunommts

### DIFF
--- a/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogButtons.kt
+++ b/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogButtons.kt
@@ -2,6 +2,7 @@ package org.nypl.simplified.ui.catalog
 
 import android.content.Context
 import android.content.res.ColorStateList
+import android.util.TypedValue
 import android.view.Gravity
 import android.view.View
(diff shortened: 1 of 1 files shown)


ray-lee (approved):
Looks good to me. Thanks for the quick fix.

PR #45, opened 2021-10-06
Use a version catalog for dependencies

io7m:
**What's this do?**
This introduces a Gradle [version catalog](https://docs.gradle.org/current/userguide/platforms.html#sub:version-catalog) to centralize the declarations of dependencies across all projects.

We have many transitive dependency conflicts in the Android application, and this goes some way towards eliminating them. It's not a complete solution (for that, we need a "platform" declaration in addition to the catalog).

Check that the application isn't broken with the updated dependencies. The upstream project isn't broken, so we should be fine. I think the A1QA suite should catch anything pretty quickly.

None.

The README was updated to indicate that `git clone --recursive` is mandatory now.

@io7m tried opening the app.

### DIFF
--- a/simplified-tests/build.gradle
+++ b/simplified-tests/build.gradle
@@ -94,6 +94,8 @@ dependencies {
   implementation libs.nypl.readium
   implementation libs.nypl.readium.js
   implementation libs.okhttp3.mockwebserver
+  implementation libs.r2.lcp
+  implementation libs.r2.shared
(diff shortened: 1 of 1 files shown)


io7m:
Fixing test failures. :weary:

io7m:
Same goes: I'm not going to merge this until I've fixed the test failures. Will hopefully be able to look at these this week.

PR #58, opened 2021-10-29
Bugfix/fix buttons inconsistency

nunommts:
**What's this do?**
This PR does the following changes:

1. Increases the buttons container height;
2. Changes the "Download" button text to "Get";
3. Adds invisible buttons to each side of a "Get" button on the book details page
4. Changes the "Cancel reservation" button text to "Remove";

Regarding the above section, we are doing these changes to:
1. Fix this text being cropped when the [book's reserve can be cancelled](https://user-images.githubusercontent.com/79104027/139486855-d02e9e0f-fb0d-4bff-a9f2-37086ed63e6c.png) or [there's a loan duration for the book](https://user-images.githubusercontent.com/79104027/139486819-c16b1236-4349-4961-bdb7-4446af17309b.png)
**UPDATE** - This error no longer happens because that string was updated, but this fix can be used for future longer texts in the button

2. Fix the inconsistency between Android and iOS in this action button, as reported [here](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=c5596de5b57d4027bf2984acfa70f930&p=fb1084c7928e4f9e8db55a42f1d2b3f4)
3. Prevent the button to fill the entire screen width as we can see [here](https://user-images.githubusercontent.com/79104027/139487023-1c777549-0e37-4de0-b67f-a022bf85e4ad.png)
4. Fix the inconsistency between Android and iOS in this action button, as reported [here](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=c5596de5b57d4027bf2984acfa70f930&p=dd41e02a85ca4b3ea5cd3adfb1c070f2)

Open the app

- Choose Lyrasis library
Search for the audiobook "A Little Princess" and reserve it.
Verify the button now says "Remove" instead of "Cancel reservation"
 - Search for the book "Gone Girl" from Gillian Flynn and get it on loan
Verify the loan duration is not cropped on the button
 - Choose PalaceBookshelf library
Open any book
Verify the button says "Get" instead of "Download"

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -50,7 +50,7 @@
         <c:change date="2021-09-10T00:00:00+00:00" summary="Introduced a new color scheme for the application."/>
       </c:changes>
     </c:release>
-    <c:release date="2021-10-26T21:54:28+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.3">
+    <c:release date="2021-10-29T18:42:54+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.3">
       <c:changes>
         <c:change date="2021-07-14T00:00:00+00:00" summary="Improve fixed-layout EPUB handling.">
           <c:tickets>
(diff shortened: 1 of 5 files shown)


ray-lee (approved):
Looks good. Thank you!

PR #59, opened 2021-11-03
Do not choke on malformed last-read locations

io7m:
**What's this do?**
This adds a catch statement to avoid crashing the app if a last-read
location is unreadable. There are likely more instances of this kind
of behaviour in the app, but we have at least fixed this one.

Affects: https://www.notion.so/lyrasis/Android-app-crashes-on-launch-and-removes-LYRASIS-Reads-library-0613d7f6e08241ea8048dbc2041fe477

Try the reproduction steps in the ticket and see if the problem re-occurs.

None.

Yes.

N/A

@io7m Tried reproducing the original issue, and added a new test case using a snapshot of an account that would trigger the problem.


dwilcox:
> **Why are we doing this? (w/ JIRA link if applicable)** Affects: https://www.notion.so/lyrasis/Error-code-51000-An-error-occurred-during-playback-on-Android-Audiobooks-db0d12b05e8f4d269011a776224b049a

I think the relevant ticket is actually: https://www.notion.so/lyrasis/Android-app-crashes-on-launch-and-removes-LYRASIS-Reads-library-0613d7f6e08241ea8048dbc2041fe477.

io7m:
Oops, yes, mis-pasted that one.

ray-lee (approved):
Looks good!

PR #60, opened 2021-11-03
Added "Return" option to all loaned books

nunommts:
**What's this do?**
This PR adds a "Return" button option to a book when it's loaned but not downloaded because of some error or status inconsistency. This PR also changes a button text from "Download" to "Get" on the catalog page in order to be consistent with what we seen in the book details screen.

There's a [reported feature](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=c5596de5b57d4027bf2984acfa70f930&p=ccbe812c04924653b14877a4f567847f) that's asking for this option so the user doesn't have to download the book before returning it.

Open the app
Choose the "Lyrasis" library
Choose any book and press the "Get" option. As soon as the download percentage shows up, close the app
Reopen the app
Verify you have the "Get" option [here](https://user-images.githubusercontent.com/79104027/140122432-148d06e2-e50c-4479-9858-aabb41715081.png) and [here](https://user-images.githubusercontent.com/79104027/140123798-1cf70d31-fa34-46c5-b9d4-b98b356b5e3c.png)
Press the "Get" button and wait.
See the book is not on your bookshelf anymore

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogBookDetailFragment.kt
+++ b/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogBookDetailFragment.kt
@@ -576,7 +576,11 @@ class CatalogBookDetailFragment : Fragment(R.layout.book_detail) {
         }
     }
 
-    val isRevocable = this.viewModel.bookCanBeRevoked && this.buildConfig.allowReturns()
+    val isRevocable = (
+      bookStatus is BookStatus.Loaned.LoanedNotDownloaded ||
+        this.viewModel.bookCanBeRevoked
+      ) && this.buildConfig.allowReturns()
+
     if (isRevocable) {
       this.buttons.addView(this.buttonCreator.createButtonSpace())
       this.buttons.addView(
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -115,9 +115,10 @@
         <c:change date="2021-10-29T00:00:00+00:00" summary="Change 'Download' button label to 'Get'"/>
       </c:changes>
     </c:release>
-    <c:release date="2021-11-03T15:24:55+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.4">
+    <c:release date="2021-11-03T16:18:33+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.4">
       <c:changes>
-        <c:change date="2021-11-03T15:24:55+00:00" summary="Fix a startup crash involving malformed bookmarks"/>
+        <c:change date="2021-11-03T00:00:00+00:00" summary="Fix a startup crash involving malformed bookmarks"/>
+        <c:change date="2021-11-03T16:18:33+00:00" summary="Added return option to all loaned books"/>
       </c:changes>
     </c:release>
   </c:releases>
--- a/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogButtons.kt
+++ b/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogButtons.kt
@@ -103,7 +103,7 @@ class CatalogButtons(
     loanDuration: String,
     onClick: () -> Unit
   ): LinearLayout {
-    return createButtonWithDuration(loanDuration, R.string.catalogDownload, onClick)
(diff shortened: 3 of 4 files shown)


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogBookDetailFragment.kt
```diff
@@ -576,7 +576,11 @@ class CatalogBookDetailFragment : Fragment(R.layout.book_detail) {
         }
     }
 
-    val isRevocable = this.viewModel.bookCanBeRevoked && this.buildConfig.allowReturns()
+    val isRevocable = (
+      bookStatus is BookStatus.Loaned.LoanedNotDownloaded ||
```
ray-lee, line 580: @nunommts Would it make sense to move this check into the `bookCanBeRevoked` function? That function is also checking the book status.

ray-lee:
@io7m Could you take a look at this one? It seems so simple that I'm afraid we're missing something.

io7m:
Yes, unfortunately, this was too simple.

The "return" process comprises a number of steps. When a book happens to be ACS-encrypted, you have to pass the "rights" file that you get during fulfillment of the book to the Adobe API. The Adobe API does a lot of nonsense in the background and then tells you if the return succeeded or not. If the return succeeded, you go on to the next step which is telling the CM that the return succeeded. The CM then deletes its own record of there being a loan.

Unfortunately, I think the above changes will allow you to get to here (for a book that's loaned but not downloaded):

https://github.com/ThePalaceProject/android-core/blob/main/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookRevokeTask.kt#L405

Because there was nothing downloaded, there won't be a rights file, and the return will trivially succeed... Except that Adobe will still think you have a loan, at least until the loan expires on its own.

jonathangreen:
@io7m Is this always the case for unfulfilled ACS encrypted items? The CM does more then just delete the loan from its DB when you call the revoke endpoint. It does make an effort to call the vendor API to return the item, but I don't know if this will work in the case of ACS items.

io7m:
I think this is a major problem we've always had with ACS - the client has to do a lot of things that the server can't. In an ideal world, for example, the server is the one that would fulfill the ACSM.

jonathangreen:
I haven't tested all the cases, I would guess this would vary by vendor, so maybe more testing is needed. But I just did a bit of local testing with Overdrive and ACS items.

Here is my setup:
I have a local copy of the circulation manager running, and I am just directly calling the circulation manager APIs, without the app being involved. The title I chose shouldn't matter, but I was using `bf635d7c-3f0b-43fc-ac5b-ab527aca77a5`.

## Test 1

- Make `get` request to `works/Overdrive%20ID/bf635d7c-3f0b-43fc-ac5b-ab527aca77a5/borrow`
- At this point the title is checked out as far as Overdrive is concerned, but not fulfilled
- Verify this is the case by directly calling the [Overdrive API](https://developer.overdrive.com/apis/checkouts) to see loan status `v1/patrons/me/checkouts`
- Make `get` request to `loans/38110/revoke`
- After this call, both the CM and Overdrive APIs agree that the loan has been returned. 

## Test 2

- Make `get` request to `works/Overdrive%20ID/bf635d7c-3f0b-43fc-ac5b-ab527aca77a5/borrow`
- Make `get` request to `works/38110/fulfill/1` to get ACSM file
- At this point the loan is "locked in" to Adobe and checked out from the Overdrive API perspective
- Verify this is the case by directly calling the [Overdrive API](https://developer.overdrive.com/apis/checkouts) to see loan status `v1/patrons/me/checkouts`
- Make `get` request to `loans/38110/revoke`
- After this call, both the CM and Overdrive APIs agree that the loan has been returned.

So at least in these cases it seems that calling the revoke endpoint in the CM is enough to return the loan. Even if the ACSM file has been requested. Whats harder for me to test is what happens if we try to call revoke if the mobile app has started to act on the ACSM file, but that fails. I'm not sure if the CM will still be able to revoke the loan in this case.

ray-lee:
Thanks for the testing, @jonathangreen. I think I'm satisfied that this works for the common case, and is an improvement over the current situation, so I'm fine with merging it as is.

ray-lee (approved):

PR #61, opened 2021-11-08
Changed user redirect screen to a book details screen after logging in from there

nunommts:
**What's this do?**
This PR adds logic to better handle the way the app behaves after an user logs in to a library after performing an action on a book details screen. There's a new event that will specify if the user is requesting a login from a book details screen so the app can intercept it and redirect the user to the catalog tab (where the book details screen is still open) after successfully logging in.

There was a [task created on Notion](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=c5596de5b57d4027bf2984acfa70f930&p=8f17f0a85cd34af09c323397f2e993ea) saying that it could be misleading and a bad UX if the user returned to the catalog page after logging in from a book details screen without knowing if the performed action was happening or not.

1. Open the app and select the Lyrasis library
2. If you're logged in, then log out from the library
3. Go to catalog
4. Request book
5. Log in
6. Verify the app is returning to the book details page and not the catalog

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -115,9 +115,10 @@
         <c:change date="2021-10-29T00:00:00+00:00" summary="Change 'Download' button label to 'Get'"/>
       </c:changes>
     </c:release>
-    <c:release date="2021-11-03T15:24:55+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.4">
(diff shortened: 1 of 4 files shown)


ray-lee (approved):
Looks good to me!

PR #62, opened 2021-11-09
Fixed Palace books not being removed from 'My Books' when they're returned

nunommts:
**What's this do?**
This PR changes the action performed on the BookRegistry class after a book is deleted: instead of calling the update for a new status, it calls the `clearFor` method that deletes the registry from the current books list and then, after the book is successfully removed, refreshes the feed. This way we prevent the books from staying in the "My Books" tab after being returned when we're using a free library.

[A bug was reported](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=c5596de5b57d4027bf2984acfa70f930&p=affd63a6978248c8a24437e07d6e7855) saying that the books were being returned but they kept appearing in the "My Books" tab which was misleading for the user.

1. Open the app and select the Palace Bookshelf library
2. Download a title
3. Go to "My Books" tab and see the book is there
4. Delete/Return the title either on the "My Books" screen or on the book details screen
5. Verify the book is not in the "My Books" tab anymore

No

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileFeedTask.kt
+++ b/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileFeedTask.kt
@@ -282,20 +282,29 @@ internal class ProfileFeedTask(
         values.filter { book -> book.book.account == accountID }
       } else {
         values
-      }.filter {
-        this.accountIsLoggedIn(it.book.account)
+      }.filter { bookStatus ->
+        if (!this.requiresLogin(bookStatus.book.account)) {
+          bookStatus.status is BookStatus.Loaned.LoanedDownloaded
+        } else {
+          this.accountIsLoggedIn(bookStatus.book.account)
+        }
       }
     return ArrayList(allBooks)
+  }
+
+  private fun requiresLogin(accountID: AccountID): Boolean {
+    return try {
+      val account = this.profiles.profileCurrent().account(accountID)
+      account.provider.authentication.isLoginPossible
+    } catch (e: Exception) {
+      false
+    }
   }
 
   private fun accountIsLoggedIn(accountID: AccountID): Boolean {
     return try {
       val account = this.profiles.profileCurrent().account(accountID)
-      if (!account.provider.authentication.isLoginPossible) {
-        true
-      } else {
-        account.loginState is AccountLoginState.AccountLoggedIn
-      }
+      account.loginState is AccountLoginState.AccountLoggedIn
     } catch (e: Exception) {
       false
     }
--- a/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileFeedTask.kt
+++ b/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileFeedTask.kt
@@ -15,6 +15,7 @@ import org.nypl.simplified.feeds.api.FeedFacet.FeedFacetPseudo.FilteringForAccou
 import org.nypl.simplified.feeds.api.FeedFacet.FeedFacetPseudo.Sorting
 import org.nypl.simplified.feeds.api.FeedFacet.FeedFacetPseudo.Sorting.SortBy
 import org.nypl.simplified.feeds.api.FeedSearch
+import org.nypl.simplified.opds.core.OPDSAvailabilityOpenAccess
 import org.nypl.simplified.profiles.controller.api.ProfileFeedRequest
 import org.nypl.simplified.profiles.controller.api.ProfilesControllerType
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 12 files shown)


simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileFeedTask.kt
```diff
@@ -282,20 +282,29 @@ internal class ProfileFeedTask(
         values.filter { book -> book.book.account == accountID }
       } else {
         values
-      }.filter {
-        this.accountIsLoggedIn(it.book.account)
+      }.filter { bookStatus ->
+        if (!this.requiresLogin(bookStatus.book.account)) {
+          bookStatus.status is BookStatus.Loaned.LoanedDownloaded
```
io7m, line 287: I'm not sure I understand why `LoanedDownloaded` specifically. Why not `Loaned`?

nunommts: If it was just `Loaned`, it would have the same effect as it was before. I filtered the `LoanedDownloaded` books for the free accounts (the ones that don't require login) because if you deleted a book it would stay on your "My Books" tab until you restarted the app. In this case, it will just show the downloaded books for this type of accounts and won't show the other ones.

I may be looking at this the wrong way, but I performed some tests and it seemed fine to me!

io7m:
I tentatively approve. I'm not quite convinced, but... if you say it works! :laughing:

nunommts:
> I tentatively approve. I'm not quite convinced, but... if you say it works! 😆

@io7m ahah let me try explain it betteer then. The condition I replaced was doing this:

`filter {
        this.accountIsLoggedIn(it.book.account)
 }`
 
 This filtering condition would display all the books if the user was authenticated in a library that requires login or display all the books if the library doesn't require login. So, if you went to the Palace bookshelf and remove a book from your books list, it would move to the state `LoanedNotDownloaded` and it would still be shown because the local feed fetching would return it anyway.

Now, with my changes it won't be displaying the `LoanedNotDownloaded` books anymore for the libraries that don't require login, because I filter them after checking if the library requires authentication on not. 

I may be saying a lot of rubbish, so please feel free to correct me if I'm wrong :D

ray-lee:
I think Nuno needs to specify `LoanedDownloaded` because an open access book is automatically considered to be `LoanedNotDownloaded`: https://github.com/ThePalaceProject/android-core/blob/0b2de99ba5355a55975a9779b599cee7f11452ab/simplified-books-registry-api/src/main/java/org/nypl/simplified/books/book_registry/BookStatus.kt#L402-L420

But that brings up a different concern: I think the check on the library not requiring login should really be a check on the book being open access. Palace Bookshelf doesn't require login, which implies that the books in it are all open access, but a library that does require login could also contain open access books. It's the fact that the book is open access (and therefore always considered to be `Loaned`) that's the real issue, not that the library doesn't require login.

nunommts:
> I think Nuno needs to specify `LoanedDownloaded` because an open access book is automatically considered to be `LoanedNotDownloaded`:
> 
> https://github.com/ThePalaceProject/android-core/blob/0b2de99ba5355a55975a9779b599cee7f11452ab/simplified-books-registry-api/src/main/java/org/nypl/simplified/books/book_registry/BookStatus.kt#L402-L420
> 
> But that brings up a different concern: I think the check on the library not requiring login should really be a check on the book being open access. Palace Bookshelf doesn't require login, which implies that the books in it are all open access, but a library that does require login could also contain open access books. It's the fact that the book is open access (and therefore always considered to be `Loaned`) that's the real issue, not that the library doesn't require login.

That was basically my main issue regarding the development of this bugfix. I did try getting three books from Palace bookshelf  and after returning one of them, it would disappear because I was filtering them to not display the `LoanedNotDownloaded` ones. So I think if you'd press "Get" they'd become `LoanedDownloaded` and if you'd press "Delete", they'd become `LoanedNotDownloaded`.

I thought it was the best way of doing this but it could cause some issues in other libraries maybe. @ray-lee Is there any way of checking what you suggested?

ray-lee:
@nuno Yes, `book.entry.availability` will be `OPDSAvailabilityOpenAccess`. Examples [here](https://github.com/ThePalaceProject/android-core/blob/0b2de99ba5355a55975a9779b599cee7f11452ab/simplified-books-registry-api/src/main/java/org/nypl/simplified/books/book_registry/BookStatus.kt#L363) and [here](https://github.com/ThePalaceProject/android-core/blob/0b2de99ba5355a55975a9779b599cee7f11452ab/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookRevokeTask.kt#L214).

nunommts:
> @nunommts Yes, `book.entry.availability` will be `OPDSAvailabilityOpenAccess`. Examples [here](https://github.com/ThePalaceProject/android-core/blob/0b2de99ba5355a55975a9779b599cee7f11452ab/simplified-books-registry-api/src/main/java/org/nypl/simplified/books/book_registry/BookStatus.kt#L363) and [here](https://github.com/ThePalaceProject/android-core/blob/0b2de99ba5355a55975a9779b599cee7f11452ab/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookRevokeTask.kt#L214).

Thank you! :) I've commited some changes and updated the PR

ray-lee:
Now that I think about this more, I have another question. Currently, if I delete a book that I downloaded from Palace Bookshelf, and I restart the app, then the deleted book doesn't show up any more. Why is that? Is the book database being updated when the app is closed or opened, but not immediately after you do the delete? It feels like we should try to get to that same state, if possible. I want to be sure we're not just hiding a deeper problem with this fix.

nunommts:
I did some debugging and the books were still being retrieved after being deleted (but with a different state). I added some breakpoints and the file deletion seemed to be working because the `file.exists()` is returning false, so I wasn't sure why it was heppening. I'll take a look.

nunommts:
@ray-lee @io7m New commit and PR description :)

io7m:
> Currently, if I delete a book that I downloaded from Palace Bookshelf, and I restart the app, then the deleted book doesn't show up any more. Why is that? Is the book database being updated when the app is closed or opened, but not immediately after you do the delete?

This is the "book registry". It's the run-time registry that contains the status of all books with which the user has some sort of relationship, and the contents of the registry aren't persisted after an app restart. It used to be that when a book was deleted, that ended the user's relationship with the book and the book status was removed from the registry, but we found that introduced a lot of unpleasant complexity when it came to revoking loans and the like. Currently, books just stay in the registry until the app is restarted. It's silly, but harmless.

io7m:
I think this PR is correct, but the semantics of book status values are quite complicated, and I could be wrong!

ray-lee (approved):
I think this looks good, based on my mental model of how this stuff works... which could be wrong.

PR #63, opened 2021-11-11
Aggressively refactor the bookmark service

io7m:
**What's this do?**
This refactors the implementation of the bookmark service to much
more aggressively save and restore bookmarks both locally and
to/from the server. Bookmarks are now regularly synced for all
enabled accounts, syncing exhaustively sends and receives all
bookmarks, and the following actions will trigger a sync:

  * Enabling syncing for an account.
  * Selecting a profile (essentially the same as opening the
    app in Palace builds).
  * Logging in to an account.
  * The app is left open for an hour. Repeats hourly.

The old functional policy implementation was removed due to
being overengineered and incomprehensible. The new implementation
is API-compatible, and all of the current unit tests pass.

Affects: https://www.notion.so/lyrasis/Palace-does-not-remember-place-in-book-or-audiobook-between-devices-on-Android-be7b4d5f3351465ca1f90ff480c6ad16

Enable bookmark syncing for an account on multiple devices, and try reading a few books. Check to see if you're receiving bookmarks from the server.

We might want to add more conditions that trigger a sync. For example, opening the bookmarks screen might be a good time to trigger a sync.

None.

Not yet; unsure if these changes are enough.

N/A

@io7m Tried syncing bookmarks rather a lot.


io7m:
Hm, seems like a couple of the tests are fragile. Looking into it...

simplified-reader-bookmarks/src/main/java/org/nypl/simplified/books/reader/bookmarks/internal/RBService.kt
```diff
@@ -0,0 +1,320 @@
...
+      when (event.state) {
+        is AccountLoggedIn ->
+          this.onAccountLoggedIn()
+
+        is AccountLoggingIn,
+        is AccountLoggingInWaitingForExternalAuthentication,
+        is AccountLoggingOut,
+        is AccountLoginFailed,
+        is AccountLogoutFailed,
+        AccountNotLoggedIn -> {
```
nunommts, line 106: Isn't a `is` missing here?

nunommts (changes requested):
I'm just not sure about that condition in the `when` flow. Other than that, and "ignoring" the failing tests, the PR looks good!

simplified-reader-bookmarks/src/main/java/org/nypl/simplified/books/reader/bookmarks/internal/RBService.kt
io7m: `AccountNotLoggedIn` is actually an `object`, so you can't refer to it in an `is` expression. Kotlin oddity.

nunommts: Got it! I thought it was a state too

io7m:
Currently working on fixing the tests. The tests need to be rewritten, frankly. The degree of mocking used just obscures the intent of the tests entirely.

io7m:
I'm having immense difficulty getting the tests to work reliably. There's some kind of race condition that's causing the tests to sporadically fail. It may be that we have to test for slightly weaker conditions now, such as "at least these requests occurred" rather than "exactly these requests occurred in exactly this order".

io7m:
I've come to realize that most of the fragility is due to this:

https://github.com/ThePalaceProject/android-core/blob/main/simplified-tests/src/test/java/org/nypl/simplified/tests/books/reader/bookmarks/ReaderBookmarkServiceContract.kt#L128

```
    this.server.enqueue(
      MockResponse()
        .setResponseCode(200)
        .setBody(response)
    )
```

Essentially, rather than being able to enqueue a response for a specific endpoint, there appears to just be a single queue that returns a response for whatever URI is requested next. This makes the tests very sensitive to timing or when operations can be reordered. I'm looking into whether `MockWebServer` supports only returning specific responses for specific endpoints.

io7m:
There is a way. I need to write a dispatcher:

https://github.com/square/okhttp/tree/master/mockwebserver#dispatcher

nunommts (approved):

io7m:
This was unnecessarily horrible. Merging!

PR #64, opened 2021-11-12
Merge upstream changes and integrate dark mode

ray-lee:
**What's this do?**

Merges changes from NYPL (https://github.com/NYPL-Simplified/Simplified-Android-Core), and integrates the dark mode support into Palace. To achieve this, simplified-ui-neutrality is refactored to inherit defaults from NYPL's Simplified-Android-Theme module, so it now just contains overrides for the attributes we want to change.

Video of dark mode:
https://user-images.githubusercontent.com/1395885/142284884-97f42bd7-d9f7-49c9-9e10-9b0d6bfa0635.mp4

Notes:
- When you build locally, you now need to set the property `org.thepalaceproject.s3.depend=true` in your gradle.properties.
- I mostly accepted changes to the README and RELEASING docs from upstream, but some of the information doesn't apply. We already had a lot of other stuff in those files that needed to be updated, so I'll do a full pass over them in a future PR.
- Dark mode doesn't yet work in the audiobook player and epub (r2) reader, since those are separate modules. I'll update those in a future PR.

To simplify the review, you can ignore changes in these directories:
- simplified-app-openebooks
- simplified-app-simplye
- simplified-cardcreator

This gets the dark mode support that was implemented upstream (along with other changes that we don't really care about).

Notion: https://www.notion.so/lyrasis/Add-support-for-dark-theme-06d7c1ac8abd438fb6c1817f0885102d

- Switch your device to dark mode, and use the app. Everything (except for the audiobook player and readium reader) should change to (mostly) have light text on a dark background. There should not be any unreadable text, buttons, or icons.
- Check that in normal (light) mode, the app still looks the same.
- Test various changes we've made from the NYPL app. Everything should still work.

None

Yes

n/a

@ray-lee checked dark mode styling throughout the app, and tested Palace features.


io7m:
This looks like a Herculean effort.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/AlternateAddressFragment.kt
```diff
@@ -61,15 +58,11 @@ class AlternateAddressFragment : Fragment(), AdapterView.OnItemSelectedListener
...
-      addressType = AlternateAddressFragmentArgs.fromBundle(it).addressType
-      if (addressType == AddressType.WORK) {
-        binding.headerTv.text = getString(R.string.work_address)
-      } else {
-        binding.headerTv.text = getString(R.string.school_address)
-      }
+    if (bundle.addressType == AddressType.WORK) {
+      binding.headerTv.text = getString(R.string.work_address)
+    } else {
+      binding.headerTv.text = getString(R.string.school_address)
```
nunommts, line 64: You can move the `binding.headerTv.text` outside the `if`  condition and then you'd just be specifying the string resource in there

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/JuvenilePolicyFragment.kt
```diff
@@ -95,129 +92,98 @@ class JuvenilePolicyFragment : Fragment() {
...
-        dialogBuilder.setMessage("$it Error")
-          .setCancelable(false)
-          .setPositiveButton(getString(R.string.try_again)) { _, _ ->
-            getToken()
+  private fun showLoading(loading: Boolean) {
+    if (loading) {
+      binding.progress.visibility = View.VISIBLE
+    } else {
+      binding.progress.visibility = View.GONE
+    }
```
nunommts, line 119: This can be simplified with the Kotlin core method by using `binding.progress.isVisible = loading`

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
```diff
@@ -852,11 +852,14 @@ class CatalogFeedFragment : Fragment(R.layout.feed), AgeGateDialog.BirthYearSele
...
       button.text = facet.title
-      button.setTextColor(ContextCompat.getColor(this.requireContext(), R.color.neutralColorPrimary))
+      button.setTextColor(
+        ContextCompat.getColor(this.requireContext(), R.color.simplified_button_text)
+      )
       button.setOnClickListener {
         this.logger.debug("selected entry point facet: {}", facet.title)
         this.viewModel.openFacet(facet)
       }
+      button.setPadding(0, 0, 0, 0)
```
nunommts, line 862: I think you can use the `setPadding(0)` from the Kotlin utils to simplify this.

nunommts (changes requested):
Just a few comments to simplify some things

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
ray-lee: Thanks. I've updated this.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/AlternateAddressFragment.kt
ray-lee: Good call, but I'd rather leave this as is, because it came from upstream, and we don't use the `cardcreator` module in Palace. If I change it, it will invite conflicts in future merges, without gaining much.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/JuvenilePolicyFragment.kt
ray-lee: Another good call, but same as above. The conflicts are a real pain when I merge in upstream changes.

nunommts (approved):

PR #65, opened 2021-11-16
Updated feed catalog ui buttons

nunommts:
**What's this do?**
This PR refactors the logic of the options in feed catalog groups regarding their UI. Instead of having a specific background for each view, we now have a background for the groups container that will handle the rounded corners and background color and each view share the same background selector: white rounded when selected and none when it's not selected.

There's a need to match both platforms UI because the iOS version handled this in a different way, as we can see [here](https://www.notion.so/lyrasis/Update-Android-UI-to-match-iOS-and-implement-new-icon-and-Brand-design-f504c6b7f5fe40c6a66c40a6e1c3438b#d36a89e20c8e498c8f7ea4c29c1303c1)

Open the app
Go to the catalog tab
See the [new UI for the groups](https://user-images.githubusercontent.com/79104027/141970203-62957c6b-2473-4cdc-9a96-ef1e1a8899c4.png)
Click on "eBooks"
Verify the "All" is not selected and the "eBooks" is

None

Yes

No

Tested by @nunommts


nunommts:
> What does this look like?

@io7m Sorry, I put it on the PR description but it's a bit "hidden" :D

device-2021-11-16-103317

io7m:
Ah, nice!

io7m (approved):
Looks good!

ray-lee (approved):
Looks good. This will probably get changed a little when I merge it into #64, because of the theme/styling refactoring that happened. I'll merge it into main first, then deal with it in an update to that PR.

PR #67, opened 2021-12-01
Added 3-step tutorial to app's first launch

nunommts:
**What's this do?**
This PR adds a new module for the tutorial feature. The tutorial consists of three steps, where each step is a different image and the user may scroll between the images back and forward. There's also a 'close' button in every step, so the user can leave at any point. After completing the splash event, there's an _AppCache_ file that checks if the tutorial was already shown or not. If so, it moves on with the app's current logic. Otherwise, it'll show the tutorial.

A [new feature](https://www.notion.so/lyrasis/Android-Add-a-3-step-tutorial-that-appears-the-first-time-the-app-is-launched-6ced6dd5a40d4379b8f27cc223839c03) was created to insert this tutorial in order to help some users who may be less familiarized with the app.

Open the app.
Verify the tutorial is showing [step1](https://user-images.githubusercontent.com/79104027/144148575-88a53e59-7199-4a36-817b-e5525d3534aa.png), [step2](https://user-images.githubusercontent.com/79104027/144148620-869f9380-26c9-4294-b283-ab18f29b0c9d.png) and [step3](https://user-images.githubusercontent.com/79104027/144148640-80baf8a1-b2dd-4972-8f20-e457af051032.png)

Skip the tutorial and verify the app's working as expected.
Close the app and reopen it.
Verify the tutorial is not being shown

_Note: If you want to see the tutorial again, you need to uninstall and install the app again or clear the app data in the mobile phone settings_

None

Yes

No

Tested by @nunommts


ray-lee:
It looks like #66 caused tests to fail. I'm going to see if I can fix those to unblock this PR.

simplified-main/src/main/java/org/nypl/simplified/main/AppCache.kt
```diff
@@ -0,0 +1,35 @@
...
+    sharedPreferences.edit().putBoolean(KEY_SEEN_TUTORIAL, true).apply()
+  }
+
+  /**
+   * Checks if the tutorial was seen
+   */
+  fun wasTutorialSeen(): Boolean {
+    return sharedPreferences.getBoolean(KEY_SEEN_TUTORIAL, false)
+  }
+}
```
ray-lee, line 35: I would prefer these to be `setTutorialSeen(boolean)` and `isTutorialSeen()`, which are more conventional for Java/Kotlin.

ray-lee (commented):
Just one comment, otherwise looks good.

PR #68, opened 2021-12-01
Complete fix for defaulting feed dates to UTC.

ray-lee:
**What's this do?**

This completes the fix to always default to UTC when parsing dates in a feed (#66). It adds changes from the upstream fix (https://github.com/NYPL-Simplified/Simplified-Android-Core/pull/948) that were not included in #66.

Unit tests were failing after #66 was merged, because not all the changes from upstream were included. See #66 for the original motivation.

Unit tests should now pass. In the app, open some book details, and verify that the publish dates look reasonable. Also see #66. 

None

No, the changelog was updated in #66.

n/a

@ray-lee ran the Palace app.


io7m:
Sorry about this. Didn't realize I'd missed those changes out.

nunommts (approved):

PR #72, opened 2021-12-07
Fix back button on error details.

ray-lee:
**What's this do?**

This makes the back button in the toolbar on the error details screen work properly. This is a continuation of the toolbar adjustments from #34, #37, and #39. Hopefully the last!

Previously, tapping the back button on the error details screen would open the switch library menu, instead of returning to the previous screen. This made it hard for users to exit from the error details screen.

Notion: https://www.notion.so/lyrasis/Back-button-doesn-t-work-on-Error-Details-50d509d048dc4ca9a816db1735dbbe94

Do something that causes an error. As of this writing, attempting to "The Secret Language of Sisters" from LYRASIS Reads will do it. Tap the Details button under the error message to open the error details screen. On the error details screen, tap the back button in the toolbar. The app should return to the previous screen. 

None.

Yes.

n/a

@ray-lee tested the back button.

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -121,7 +121,7 @@
         <c:change date="2021-11-08T00:00:00+00:00" summary="Return to book details after logging in while borrowing a book"/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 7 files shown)


nunommts (approved):
Looks good!

PR #74, opened 2021-12-15
Disabled landscape mode

nunommts:
**What's this do?**
This PR sets the screen orientation as "Portrait" for every activity to disable the landscape mode.

There's a [task request](https://www.notion.so/lyrasis/Disable-landscape-mode-c4564cd3379e43b6be0ff0460fd5bcd6) to disable the landscape mode in order to fix some issues and to be aligned with the iOS app.

Open the app (make sure your device has the screen rotation enabled in the status bar options)
Go to every screen, rotate the device and verify the app is not rotating and the layout remains as it was

None

Yes

No

Tested by @nunommts


ray-lee (approved):
Looks good!

PR #76, opened 2021-12-20
Turn on sync bookmarks by default

nunommts:
**What's this do?**
This PR turns on the sync bookmarks option in a library after logging in to it, by changing the default value to be returned for that option when it wasn't manually set by the user.

There's a [feature request](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=c5596de5b57d4027bf2984acfa70f930&p=14179c6f09ca4b72a18bdf4b136f7ef5) to change this in order to provide a better experience for the user.

Uninstall the app or clear its data
Open the app
Select "Lyrasis" library
Login
Verify the 'Sync bookmarks' option [is on by default](https://user-images.githubusercontent.com/79104027/146758635-fb15b5b2-dae1-4b06-ab9b-7a7a6ac4d48b.png)
Turn off the option
Logout
Login
Verify the 'Sync bookmarks' option is off by default

None

Yes

No

Tested by @nunommts


ray-lee (commented):
@nunommts It's not clear to me why the changes to RBServiceOpCheckSyncStatusForAccount.kt are needed. Can you explain what they do?

nunommts:
@ray-lee The _runActual()_ method on RBServiceOpCheckSyncStatusForAccount.kt was setting the default value to _false_ the first time it was being called, even if we were setting the default value to true in other places, because it was basically checking if the returned value was null or not. It seemed unnecessary to me and I left the other parts of the code for logging purposes.

ray-lee:
@nunommts I'm thinking about what that code would be used for. I think it's syncing the setting, right? So if on one device, you toggle the Sync Bookmarks setting, other devices will check the server for an updated setting before trying to sync. Maybe the idea is that the Sync Bookmarks setting is associated with the account, not with any particular device, so if you change it on one device, it should propagate to other devices with the same account. So we might be breaking that if we remove that code.

Instead, maybe we should change it so that if no Sync Bookmarks setting exists the server, it defaults to true instead of false. It looks like that would happen on these lines:
https://github.com/ThePalaceProject/android-core/blob/d8f4982ab04b9d8462ae76fb37732f874899cb29/simplified-reader-bookmarks/src/main/java/org/nypl/simplified/books/reader/bookmarks/internal/RBHTTPCalls.kt#L207
https://github.com/ThePalaceProject/android-core/blob/d8f4982ab04b9d8462ae76fb37732f874899cb29/simplified-reader-bookmarks/src/main/java/org/nypl/simplified/books/reader/bookmarks/internal/RBHTTPCalls.kt#L210

What do you think?

nunommts:
@ray-lee Nice catch! I searched all the code for this kind of default values and I thought I've covered them all and that method was the reason why it was failing. It certainly looks like that if we change these two values to true, it might be it! I'll update the code as soon as I can

ray-lee (approved):
Looks good!

PR #78, opened 2022-01-06
Sync bookmarks when opening epub reader

ray-lee:
**What's this do?**

This syncs bookmarks (and the current reading position, which is a special bookmark) from the server when a book is opened in the readium epub viewer, prior to displaying the book and navigating to the current position. Previously, bookmarks were synced when the app was started, when changing libraries, and on an hourly schedule, but not when a book was opened.

The reading position within a book should be retained between devices. Previously, if a user read a book on device 1, then opened the book on device 2, the reading position would not be the same as on device 1, unless the user waited up to an hour for the next sync, or restarted the app, or changed to a different library and back. This forces the bookmarks to sync when a book is opened, to ensure that the current reading position is correct.

Notion: https://www.notion.so/lyrasis/Palace-does-not-remember-place-in-book-between-devices-on-Android-be7b4d5f3351465ca1f90ff480c6ad16

- On two devices, start the Palace app and log in to the same library.
- On device 1, open a book. Make a few bookmarks on random pages, and turn to another random page.
- On device 2, open the same book.

Verify that on device 2, the book is opened to the same chapter and page number that is open on device 1. (If the devices have different screen sizes, the page number may be slightly different, but should be close). Verify that the bookmarks from device 1 all appear on device 2.

None.

Yes.

n/a

@ray-lee tested on the Palace app.

### DIFF
--- a/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Bookmarks.kt
+++ b/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Bookmarks.kt
@@ -31,8 +31,8 @@ object Reader2Bookmarks {
   ): ReaderBookmarks {
     return try {
       bookmarkService
-        .bookmarkLoad(accountID, bookID)
-        .get(10L, TimeUnit.SECONDS)
+        .bookmarkSyncAndLoad(accountID, bookID)
+        .get(15L, TimeUnit.SECONDS)
     } catch (e: Exception) {
       this.logger.error("could not load bookmarks: ", e)
       ReaderBookmarks(null, emptyList())
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -133,13 +133,14 @@
         <c:change date="2021-12-07T00:00:00+00:00" summary="Fixed back button not working on Error Details screen"/>
       </c:changes>
     </c:release>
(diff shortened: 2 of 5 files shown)


simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Bookmarks.kt
```diff
@@ -31,8 +31,8 @@ object Reader2Bookmarks {
   ): ReaderBookmarks {
     return try {
       bookmarkService
-        .bookmarkLoad(accountID, bookID)
-        .get(10L, TimeUnit.SECONDS)
+        .bookmarkSyncAndLoad(accountID, bookID)
+        .get(15L, TimeUnit.SECONDS)
```
nunommts, line 35: Why did this value change from 10 to 15?

nunommts (approved):
Just left a comment out of curiosity but it shouldn't be relevant. Other than that, it looks great!

simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Bookmarks.kt
ray-lee: 10 seconds seemed to be cutting it close on the emulator, when sometimes things are really slow. In real life it shouldn't matter, but I figured I'd give it some extra time, since it now has to do an additional network request.

PR #81, opened 2022-01-12
Added 'isOpenAccess' flag to Loaned.NotDownloaded book status

nunommts:
**What's this do?**
This PR adds a flag (isOpenAccess) to the Loaned.NotDownloaded book status to distinguish the possible action between open access books and books from libraries that require authentication.

This change is being made so the user's experience can be clearer in terms of actions to perform over a book. Right now the app was either showing "Get" or "Download" for the books, and this way we can show "Get" when the user doesn't have downloaded the book and it is open access and we show "Download" when the user doesn't have downloaded the book and it isn't open access.

Open the Palace app
Go to the "Palace Bookshelf" library
Open any book and [verify the button says "Get"](https://user-images.githubusercontent.com/79104027/149134247-352298e8-2ca3-44e2-a15a-9c4a592a0417.png)
Go back and switch to the "Lyrasis Reads" library (or any other that requires authentication)
Open any book, [verify the button says "Get](https://user-images.githubusercontent.com/79104027/149135439-4a257bad-8b5f-4008-b267-b21dd1a3f089.png) and press it
Wait until the book is successfully downloaded
Logout from the library, go to the device settings, open the Palace app settings and clear data
Open the Palace app
Log in to the "Lyrasis Reads" library
Go to "My Books" and [verify the previously selected book's action says "Download"](https://user-images.githubusercontent.com/79104027/149135573-09bb3684-e228-4e83-9765-d5a2efa3453a.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -133,7 +133,7 @@
         <c:change date="2021-12-07T00:00:00+00:00" summary="Fixed back button not working on Error Details screen"/>
       </c:changes>
     </c:release>
-    <c:release date="2022-01-07T22:14:49+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.6">
+    <c:release date="2022-01-12T11:19:53+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.6">
       <c:changes>
         <c:change date="2021-12-14T00:00:00+00:00" summary="Added audiobook player commands to lock screen"/>
         <c:change date="2021-12-15T00:00:00+00:00" summary="Disabled landscape mode"/>
@@ -141,7 +141,8 @@
         <c:change date="2021-12-20T00:00:00+00:00" summary="Turn on sync bookmarks by default"/>
         <c:change date="2021-12-27T00:00:00+00:00" summary="Added 'more' label to catalog list titles"/>
         <c:change date="2022-01-05T00:00:00+00:00" summary="Fixed reading position from other devices not remembered when opening a book."/>
-        <c:change date="2022-01-07T22:14:49+00:00" summary="Changed the button label for the checked out books"/>
+        <c:change date="2022-01-07T00:00:00+00:00" summary="Changed the button label for the checked out books"/>
+        <c:change date="2022-01-12T11:19:53+00:00" summary="Added isOpenAccess flag to Loaned.NotDownloaded book status"/>
       </c:changes>
     </c:release>
   </c:releases>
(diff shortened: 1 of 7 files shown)


ray-lee (commented):
@nunommts This looks fine to me, but for some reason the CI build didn't run on this PR, so I can't see test results. I wouldn't be surprised if some tests need to be updated. If there aren't any tests that cover this, I think we should add some.

nunommts:
@ray-lee It didn't run the tests because there were some merge conflicts with the main branch. I fixed them and it seems to be working. The tests didn't require any updates, but since this PR is just to add a flag for UI purposes and it has no influence on the action performed, do you think adding those tests would be useful?

ray-lee:
@nunommts My opinion is that the changes to simplified-ui-catalog don't need tests, for the reasons you stated -- they would be redundant with the appium tests that A1QA is doing. But the changes to simplified-books-borrowing and simplified-books-registry-api do, because they are API changes that could affect other things, now or in the future. We want to ensure that those changes always work as expected, because we will potentially build other functionality using them.

My general rule is that any changes outside of the simplified-ui-* modules should have associated tests. Does that make sense?

ray-lee:
As I see it, the intent of the modification to `BorrowLoanCreate` is that if a book is open access, `isOpenAccess` will be true when the state transitions to `LoanedNotDownloaded`. That doesn't necessarily have anything to do with UI. The rest of the application could now rely on that to do anything.

I would at least add a test to `BorrowLoanCreateTest`, to verify that if a book is open access, it has state `LoanedNotDownloaded` with `isOpenAccess` set to true. Hopefully you can figure out how to do that from the other tests. I would also add an assertion to at least one of the existing tests, to say that `LoanedNotDownloaded.isOpenAccess` should be false (assuming the test is for a book that is not open access).

The change to `BookStatus.fromBook` should also be tested, but I don't see _any_ existing tests for that, so I'm willing to let it go. It's kind of surprising that it's not being tested though.

ray-lee:
Also, it looks like an extraneous change to SplashFragment.kt snuck into this PR. That should be removed, because it reverts  #80.

nunommts:
@ray-lee Sure, I'll take a look at it and add one or more tests. Thanks for clarifying that. I honestly don't know how those changes returned to the SplashFragment, but I've removed them again.

ray-lee (approved):
Looks good. Thanks for adding that test!

PR #82, opened 2022-01-13
Fixed book title request infinite loading

nunommts:
**What's this do?**
This PR changes the tab where the user is coming from when entering the authentication screen for a library after trying to get a book on loan. This way, the user will stay in the settings tab if he log on to a library from there and will be redirected to the feed/book details screen after returning (logged in or not) from the auth screen (if the book required it). This PR also fixes the infinite loading bug that was happening if the user returned from the authentication screen to the book details page without logging in.

There's a [bug report indicating these two issues](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=c5596de5b57d4027bf2984acfa70f930&p=c72840a1fe73413f9f17f3a98f5a6e15) (infinite loading and not returning to feed/book details screen) and solving them will make the user experience less confusing and will also make it possible for the user to retry an operation over a book even if he didn't log in to the library.

Open the Palace app
Select "Lyrasis Reads" library
Logout from the library if you're logged in
Open any book from the catalog
Click on "Get"
In the auth screen, go back and [confirm you're in the book details page with the "Get" button visible and enabled](https://user-images.githubusercontent.com/79104027/149379026-48235681-3ad5-48ae-8598-b3a4caf1d257.png)
Click on "Get" again
In the auth screen, insert your credentials and wait
Return to the book page and confirm [the download has started](https://user-images.githubusercontent.com/79104027/149379084-e71b29d9-0994-434d-b695-94b51d32221d.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -133,7 +133,7 @@
         <c:change date="2021-12-07T00:00:00+00:00" summary="Fixed back button not working on Error Details screen"/>
       </c:changes>
     </c:release>
-    <c:release date="2022-01-11T23:16:46+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.6">
+    <c:release date="2022-01-13T15:59:25+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.6">
       <c:changes>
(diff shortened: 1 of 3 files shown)


ray-lee (approved):
Looks good!

PR #85, opened 2022-01-20
Fixed crash when opening the TOC page on audiobooks

nunommts:
**What's this do?**
This PR changes the way the _AudiobookPlayerActivity_ handles the switch from the _PlayerFragment_ to the _TOCFragment_. Instead of replacing the first one, the activity hides it which will prevent it from being destroyed and to destroy the PlayerService. In this way, the behavior remains the same and the app doesn't crash when opening the TOC screen.

There's a [crash report](https://www.notion.so/lyrasis/Android-Tapping-in-ToC-for-audiobook-causes-app-to-crash-297b9396a30544ccafcd17174f3eba2f) saying that this was happening in some audiobooks, but the error could be most likely happening on every audiobook. By fixing this, the app doesn't crash anymore when the user opens the TOC screen.

In order to reproduce the exact steps of the crash report:
Open the Palace app
Go to settings and press 7 times the commit version to enable the debug options
Open the debug options and enable the "Show Non-Production libraries" option.
Go to the libraries screen and choose `Overdrive Integration Test Library`
Checkout the audiobook  *Inferno* by Dan Brown
Start listening to that audiobook
Open the table of contents (ToC)
Confirm the app doesn't crash
Select any chapter of the ToC
Confirm the app doesn't crash

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -133,7 +133,7 @@
         <c:change date="2021-12-07T00:00:00+00:00" summary="Fixed back button not working on Error Details screen"/>
       </c:changes>
     </c:release>
-    <c:release date="2022-01-19T12:00:22+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.6">
+    <c:release date="2022-01-20T14:44:02+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.6">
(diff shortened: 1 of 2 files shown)


ray-lee (approved):
Nice!

PR #86, opened 2022-01-21
Removed Show filter

nunommts:
**What's this do?**
This PR removes the 'Show' filter and the corresponding options from the "My Books" tab

[The "Show" filter is a bit redundant](https://www.notion.so/lyrasis/Remove-Show-filter-on-Android-17c955d0cfd64ce189ce5e81d847abd6) because the associated options ("All" and "On Loan") have no practical effect because all the books in the "My Books" tab are loaned to the user, so removing it will cause less complexity to the app.

Open the Palace app
Select "Lyrasis Reads" library
Open "My Books" tab
Verify [the "Show" filter option was removed](https://user-images.githubusercontent.com/79104027/150587137-2e382878-db97-4acb-b82b-8ca4aff7e0a9.png) and the only option at the top is the "Sort By"

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -133,7 +133,7 @@
         <c:change date="2021-12-07T00:00:00+00:00" summary="Fixed back button not working on Error Details screen"/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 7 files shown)


ray-lee (approved):
Looks good.

PR #87, opened 2022-01-25
Add CI workflow to create releases

ray-lee:
**What's this do?**

This adds a GitHub workflow to automate a few of the steps required to create a release of the app. Everything basically works the same as before, except that instead of pushing a tag to trigger the release, you now run a workflow (via the GitHub Actions screen). The workflow will:

1. Tag the release.
2. Create a GitHub Release from the tag, and put the generated release notes on the release page.
3. Build and deploy the app to the Play Store as usual, but instead of sending all the release notes to the Play Store, only generate a link that refers users to the release notes in GitHub.

The RELEASING.md documentation has been updated, and cleaned up to change NYPL-specific instructions to be relevant to Palace.

The Play Store has a limit on the size of the release notes, and we were sometimes hitting it, which would make the release fail. This should fix that. Notion: https://www.notion.so/lyrasis/Publishing-to-Google-Store-fails-because-release-notes-are-too-long-07e0a00d535c4f98b5fe5a7ef8ce44c3

When it's time for the next release, follow the instructions in RELEASING.md. Verify that:
- The Create Release workflow succeeds.
- A tag is created on the release branch.
- A release is created in GitHub (under the Releases section in the sidebar of the Code tab).
- The release page contains the release notes generated by the changelog tool.
- A draft release is created in the Play Store, and the release notes contain a link to the release notes in GitHub.

None

No, since this isn't a user-facing change.

Yes. See RELEASING.md.

@ray-lee tested this as much as possible without actually doing a release. I'll keep an eye on the next release to make sure everything is working.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -24,14 +24,6 @@ $ git submodule init
 $ git submodule update --remote --recursive
 ```
 
-#### SimplyE Or Open eBooks?
-
-Currently, this project contains two published applications: SimplyE and Open eBooks.
-The instructions here are written in a manner that allows for both applications
-to follow independent release cycles. The only parts that change in each case
-are the `gradle.properties` file that must be edited, and the names of the `git`
-tags.
-
 #### Prepare To Create A Release Branch
 
 The first step required is to set the project version numbers to the values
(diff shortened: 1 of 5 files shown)


RELEASING.md
```diff
@@ -246,25 +222,20 @@ $ git push
...
-```
+1. In GitHub, open the Actions tab, and select the "Create Release" workflow.
+2. Open the "Run workflow" dropdown.
+3. In the "Use workflow from" dropdown, select the `release/99.0.0` branch.
+4. Click the "Run workflow" button.
 
-When the CI encounters a tagged commit, it will push the build it produces to
-the Play Store, and will push various components to Maven Central.
+The workflow will tag the commit, and create a GitHub release with generated release
+notes. It produce a release build, and push it to the Play Store, with a link to the
```
nunommts, line 237: Isn't there a typo in "It produce"? I was thinking of "It produces" or "It will produce", maybe.

nunommts (approved):
I just left a comment about a possible typo. Other than that, it looks great! :)

RELEASING.md
ray-lee: Yes, thank you!

PR #88, opened 2022-02-03
Removed sync bookmarks option from libraries that don't require authentication

nunommts:
**What's this do?**
This PR adds logic to check if a library requires credentials in order to show or hide the sync bookmarks option. It also rearranges a bit the UI so, when the library doesn't require authentication, it doesn't look like [this](https://user-images.githubusercontent.com/79104027/152415534-d76fba2c-4855-45c9-b909-29400bf6deff.png). Finally, this PR also fixes the button action position in the book details page when it's reserved and can be removed, so it looks like [this](https://user-images.githubusercontent.com/79104027/152415745-a23d47a1-9625-488e-815b-ed24109d3d13.png).

There's a [ticket to remove the sync bookmark option](https://www.notion.so/lyrasis/Remove-Sync-Bookmarks-setting-for-Palace-Bookshelf-on-Android-12490cdf78da4eb9b25b18ec0c5b178c) for the libraries that don't require authentication because the user will never be able to log in to that library and to get his bookmarks synced in differente devices. The UI related fixes were made to improve the app and to make the user's experience better.

Open the app
Select the "Palace Bookshelf" library
Open settings
Verify [the 'Sync Bookmarks' option is not there](https://user-images.githubusercontent.com/79104027/152416304-3063e754-bd01-43c1-98e9-d4d96caad788.png)
Go back and select "Lyrasis Reads" library
Open settings
Verify [the 'Sync Bookmarks' option is there](https://user-images.githubusercontent.com/79104027/152416340-b6f9079f-5f0a-4bb8-a566-26434b61664e.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -151,10 +151,11 @@
         <c:change date="2022-01-21T00:00:00+00:00" summary="Removed the &quot;Show&quot; filter from My Books. This wasn't useful."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-02-03T15:19:49+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.7">
+    <c:release date="2022-02-03T19:19:14+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.7">
       <c:changes>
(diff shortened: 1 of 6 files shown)


ray-lee (approved):
Looks good!

PR #90, opened 2022-02-08
Fixed related books error

nunommts:
**What's this do?**
This PR fixes an URI parsing that was making it unparseable by the app, leading to an error message when trying to open some book related books.

There's a [bug report](https://www.notion.so/lyrasis/Error-appears-when-you-try-to-open-related-books-6b840acfd4354598b75fb38dd647e51b) indicating that the books from the Palace Bookshelf are not allowing to open their related books feed.

Open the app
Select "Palace Bookshelf" library
Open any ebook
Scroll down and click on "Related books..."
Verify the feed [is being correctly opened](https://user-images.githubusercontent.com/79104027/152983229-07382453-de59-445d-bd5d-6a5bf2d4fea1.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -151,13 +151,14 @@
         <c:change date="2022-01-21T00:00:00+00:00" summary="Removed the &quot;Show&quot; filter from My Books. This wasn't useful."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-02-08T03:20:11+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.7">
+    <c:release date="2022-02-08T11:14:45+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.7">
       <c:changes>
         <c:change date="2022-01-28T00:00:00+00:00" summary="Fixed the audiobook playback rate getting reset back to 1x after pausing."/>
         <c:change date="2022-02-03T00:00:00+00:00" summary="Fixed a crash that could happen when exiting an audiobook while it's playing."/>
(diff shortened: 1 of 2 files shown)


ray-lee (commented):
@nunommts What does the uri to the related books look like, and how was the `normalize()` call breaking it? Do you know where else the `resolveFeedFromBook` function gets used, and do those links continue to work?

nunommts:
@ray-lee The `resolveFeedFromBook` is only called by a method that is only called in the book details screen by clicking on the 'related' view, so basically it's just called when the user presses "Related Books..." on that screen.

I tried with several examples from both Palace Bookshelf and Lyrasis reads. Two examples:

*Palace Bookshelf*
`https://openbookshelf.dp.la/OB/works/URI/https://standardebooks.org/ebooks/charlotte-bronte/shirley/related_books`

*Lyrasis Reads*
`https://gorgon.tpp-qa.lyrasistechnology.org/lyrasis-reads/works/Bibliotheca%20ID/e1o42g9/related_books`


Now, after calling the `normalize()`, the URIs become, respectively:

`https://openbookshelf.dp.la/OB/works/URI/https:/standardebooks.org/ebooks/charlotte-bronte/shirley/related_books`
`https://gorgon.tpp-qa.lyrasistechnology.org/lyrasis-reads/works/Bibliotheca%20ID/e1o42g9/related_books`

As you can see, the second one didn't have any changes but the first one got one of its paths changed from _URI/https://_ to _URI/https/_ which will cause the URI parsing to fail and that's why I removed that method.

ray-lee:
Great, thanks. This looks good to me.

ray-lee (approved):

PR #93, opened 2022-02-15
Updated debug options strings

nunommts:
**What's this do?**
This PR updates one of the debug options label and creates a string resource for each debug option on the settings tab

There's a [feature request](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=22cd92c0cc5c46409134c9f380e6e718&p=2b4cee827a12476b8ce7b5166c7ece94) to make the app similar to the iOS version by changing a specific label for one of the existing debug options. Also, the string resources were created in order to properly create the debug settings layout without any hardcoded strings, which is a bad programming practice.

Open the app
Go to the Settings tab
Tap 7 times on the commit version
Open the Debug Options
Scroll down until the _Testing Features_ header and verify the first setting now says ["Enable Hidden Libraries"](https://user-images.githubusercontent.com/79104027/154166967-6633ad88-874e-4350-961c-787c0207bfc6.png) instead of "Show non-production" libraries

None

Yes

No

Tested by @nunommts


ray-lee (approved):
Looks good. Glad you moved all those strings to resources.

PR #96, opened 2022-02-24
Fix book titles being cut off on book details screen

nunommts:
**What's this do?**
This PR removes the _maxLines_ attribute in the book details screen xml layout so the title is not limited and can fully shown despite its length. It also decreases a bit the font size so the title container doesn't get too huge.

There's [a report](https://www.notion.so/lyrasis/Determine-when-long-titles-are-cut-off-on-detail-view-Android-c813fa6f75714efa9756fdf03c3d213f) stating that some titles [are being cut off](https://user-images.githubusercontent.com/79104027/155557083-24acc01a-17e0-45d8-95b0-3020fd6c4472.png) in the details page when they are too large. However, since it's the details page and that's the screen where the user could have access to the book's full information, there's a need to make it fully displayed.

Open the app
Select any library
Select any book or audiobook with a large title and open it
Verify [the title's being fully displayed](https://user-images.githubusercontent.com/79104027/155553725-99f40271-cbd4-4113-af12-f2bdc43a83ba.png)

None

Yes

No

Tested by @nunommts


ray-lee (approved):
Looks good!

PR #99, opened 2022-03-04
Decode base64-encoded LCP hashed passphrases.

ray-lee:
**What's this do?**

This adds support for base-64 encoded LCP hashed passphrases. For LCP encrypted books in the loans feed, the `lcp:hashed_passphrase` element contains the hashed passphrase to be used as a key to decrypt the book. Currently, the CM sets this value to a hex string (a sha-256 hash). According to the spec, this value should be base64 encoded. The app will now check if a hashed passphrase is base64 encoded, and decode it if necessary, before using it as a key for decryption.

The LCP automatic key retrieval spec wants the hashed passphrase to be base64 encoded in transit:
https://readium.org/lcp-specs/notes/lcp-key-retrieval.html#sample-of-readium-web-publication-manifest-supporting-a-link-to-an-lcp-license-and-an-lcp_hashed_passphrase-property

Notion: https://www.notion.so/lyrasis/Decode-base64-encoded-hashed-passphrases-Android-c9cb35a8ddbe4b22869b181be398c9f2

LCP epubs should continue to open successfully. The CM has not been updated yet to send base64-encoded hashed passphrases. Once it has, the app should be tested again to ensure that LCP epubs (both previously downloaded and new downloads) still open successfully.

Unit tests are included.

None for merging, but the CM needs to be updated to fully test this.

Yes.

n/a

@ray-lee ran this.

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -151,7 +151,7 @@
         <c:change date="2022-01-21T00:00:00+00:00" summary="Removed the &quot;Show&quot; filter from My Books. This wasn't useful."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-03-01T15:01:30+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.7">
+    <c:release date="2022-03-04T15:24:22+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.7">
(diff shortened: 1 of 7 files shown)


nunommts (approved):
It looks great to me!

PR #100, opened 2022-03-09
Add support for borrowing LCP audiobooks.

ray-lee:
**What's this do?**

This adds support for borrowing and downloading LCP audiobooks (and eventually, PDFs).

- The `BorrowLCP` borrowing subtask only allowed the LCP acquisition path to result in an EPUB. It now allows the result to be an audiobook, and leaves a stub to allow supporting PDF.
- The `application/audiobook+lcp` mime type is now recognized, so that the LCP acquisition path will be followed.

This is the first step to support borrowing and playing LCP audiobooks. Notion: https://www.notion.so/lyrasis/Get-LCP-for-audiobooks-working-in-the-Palace-Android-app-46e982c74a2949d9b6049d6212a10380#dfb4e416ff57468eac0f993a0d9e3ec9

I've added unit tests for BorrowLCP, including tests for downloading EPUB books and audiobooks.

Borrowing and downloading an LCP audiobook (from the DPLA Titles lane in LYRASIS Reads) should be successful. The book should take a relatively long time to download, because the LCP acquisition will contain all the audio files, not just the manifest.

Downloaded audiobooks will not play yet ("URI is not absolute" error message).

None.

Yes

n/a

@ray-lee downloaded some LCP audiobooks.

### DIFF
--- a/simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/DatabaseFormatHandleAudioBook.kt
+++ b/simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/DatabaseFormatHandleAudioBook.kt
@@ -17,6 +17,7 @@ import org.nypl.simplified.books.api.BookDRMKind
 import org.nypl.simplified.books.api.BookFormat
 import org.nypl.simplified.books.book_database.api.BookDRMInformationHandle
 import org.nypl.simplified.books.book_database.api.BookDatabaseEntryFormatHandle.BookDatabaseEntryFormatHandleAudioBook
+import org.nypl.simplified.files.DirectoryUtilities
 import org.nypl.simplified.files.FileUtilities
 import org.nypl.simplified.json.core.JSONParserUtilities
 import org.nypl.simplified.json.core.JSONSerializerUtilities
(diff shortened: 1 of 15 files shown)


nunommts (approved):

simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/DatabaseFormatHandleAudioBook.kt
```diff
@@ -133,7 +137,17 @@ internal class DatabaseFormatHandleAudioBook internal constructor(
...
+      } else {
+        FileUtilities.fileDelete(this.fileBook)
+      }
+
+      this.formatRef = this.formatRef.copy(
+        file = null,
+        position = null
+      )
+
       this.formatRef
```
nunommts, line 151: You can just remove the assignment and the return value in the end and just leave the

```
this.formatRef.copy(
        file = null,
        position = null
      )
```

nunommts (approved):

simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/DatabaseFormatHandleAudioBook.kt
```diff
@@ -240,6 +254,27 @@ internal class DatabaseFormatHandleAudioBook internal constructor(
...
+      }
+
+      this.formatRef = this.formatRef.copy(
+        file = this.fileBook,
+        manifest = BookFormat.AudioBookManifestReference(
+          manifestURI = URI("manifest.json"),
+          manifestFile = this.fileManifest
+        )
+      )
+      this.formatRef
```
nunommts, line 272: You can remove the assignment and the return value in the end and just leave the

```
this.formatRef.copy(
        file = this.fileBook,
        manifest = BookFormat.AudioBookManifestReference(
          manifestURI = URI("manifest.json"),
          manifestFile = this.fileManifest
        )
      )
```

PR #106, opened 2022-04-07
Add support for LCP-protected PDFs.

ray-lee:
**What's this do?**

This adds support for reading LCP-protected PDFs. These are downloaded as zip files, containing an encrypted PDF. Readium's kotlin-toolkit is used to open and decrypt the file, and pass an input stream to the reader.

Notion: https://www.notion.so/lyrasis/Add-support-for-LCP-encrypted-PDFs-Android-c14bd90d074941759ce038c62aa4a272#26d1b8829815434ab9f5dd452af46b23

In LYRASIS library, borrow and open "Spin Doctors" by Nora Loreto. The book should open successfully in the PDF reader. Other titles from the PDF Test lane may work, depending on the device, but they may be too large for the PDF reader to handle.

None

Yes

n/a

@ray-lee opened an LCP PDF.

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -174,8 +174,10 @@
         <c:change date="2022-03-22T00:00:00+00:00" summary="Changed the label of untitled audiobook files from &quot;Chapter&quot; to &quot;Track&quot;."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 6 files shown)


nunommts (approved):
LGTM!

PR #114, opened 2022-05-17
Show loading indicator while LCP passphrase is being retrieved

nunommts:
**What's this do?**
This PR forces the downloading status event to be sent before retrieving the book's passphrase when it's needed, so it can happen while a loading animation is displayed to the user instead of display the action buttons without any feedback

There's a reported bug saying that while the book's passphrase is being retrieved, the user sees the "Download" and "Return" buttons instead of a loading animation, which is misleading and causes a bad UX

Open the app
Open any book from the "DPLA Titles" section
Click on "Get"
Confirm you get [two loading animations: one for the LCP passphrase retrieval and the other for the book itself](https://user-images.githubusercontent.com/79104027/168845393-6a67f75d-1714-403d-862c-6b9fd6f587a7.mp4)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -190,7 +190,7 @@
         <c:change date="2022-04-29T00:00:00+00:00" summary="Fixed &quot;Unable to initialize audio engine&quot; error playing some audio books."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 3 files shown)


ray-lee (approved):
Simple enough!

PR #117, opened 2022-05-19
Added 'More' button to the book details screen

nunommts:
**What's this do?**
This PR adds logic to get the number of lines the book description is filling after being set and in case it's higher than 6, it sets the maximum of lines to 6 and displays a "More" button that will display the full description when it's clicked.

There's a [feature request](https://www.notion.so/lyrasis/Android-There-is-no-More-button-in-the-book-detail-view-1632eebb651447ed97bbf8e0fd55afa4) to add this logic to make the app more similar to the iOS version

Open the app
Select "Palace Bookshelf" library (it can be other library)
Open _Macbeth_ by William Shakespeare (it can be other book)
Verify the [book's description is not fully visible and there's a "More..." label below it](https://user-images.githubusercontent.com/79104027/169325603-5d0d1202-0022-491d-b8aa-c5ba844278f4.png)
Click on "More..."
Verify [the book's description is now fully visible](https://user-images.githubusercontent.com/79104027/169325680-f0edc74d-6e85-40e4-9ddc-8c27ffd53fbf.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -190,7 +190,7 @@
         <c:change date="2022-04-29T00:00:00+00:00" summary="Fixed &quot;Unable to initialize audio engine&quot; error playing some audio books."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-05-17T17:53:36+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.11">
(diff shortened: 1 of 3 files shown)


ray-lee (approved):
Looks good!

PR #123, opened 2022-07-12
Added dialog with the loan limit message

nunommts:
**What's this do?**
This PR adds a new state for a book borrowing result: the BookReachedLoanLimit. In order to assign it, we're comparing the error report type and then broadcasting the new status. Whenever the views receive this state, they'll immediately call their viewModel method to restore the book initial status so the options can be the same as they were before the book borrowing request and they'll display a dialog with the proper message.

There's a [feature request](https://www.notion.so/lyrasis/Android-There-is-an-error-instead-of-alert-when-the-limits-of-books-for-loan-are-exceeded-b3e8d307daec41d580c97733dd58a38e) to make this behavior similar to the iOS version.

Open the Palace App
Download 10 books from different distributors (or 5 books from Bibliotheca)
Open a 11th book and try to download it
Verify there's a [dialog being shown with the loan limit message](https://user-images.githubusercontent.com/79104027/178565194-bc3d10f5-e638-442b-891d-f94e8585c589.png)
Dismiss it and verify [the option is the same as it was before the request](https://user-images.githubusercontent.com/79104027/178565155-10d2668d-77bc-4d7e-9fc3-579bd2fa3197.png)
Go back to the Catalog
Click on "More..." on any lane
Try to download any book
Verify there's a [dialog being shown with the loan limit message](https://user-images.githubusercontent.com/79104027/178565252-46279d51-702e-4c96-8bdd-8f9886c103d6.png)
Dismiss it and verify [the option is the same as it was before the request](https://user-images.githubusercontent.com/79104027/178565297-3ea1ea10-7d97-4d54-9769-e244018490e8.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-accessibility/src/main/res/values/strings.xml
+++ b/simplified-accessibility/src/main/res/values/strings.xml
@@ -8,4 +8,5 @@
   <string name="bookIsDownloading">The book \'%1$s\' has started downloading</string>
   <string name="bookIsOnHold">A reservation has been placed for the book \'%1$s\'</string>
   <string name="bookReturned">The book \'%1$s\' has been successfully returned</string>
+  <string name="reachedLoanLimit">u have reached your loan limit</string>
 </resources>
--- a/simplified-accessibility/src/main/res/values/strings.xml
+++ b/simplified-accessibility/src/main/res/values/strings.xml
@@ -8,5 +8,5 @@
   <string name="bookIsDownloading">The book \'%1$s\' has started downloading</string>
   <string name="bookIsOnHold">A reservation has been placed for the book \'%1$s\'</string>
   <string name="bookReturned">The book \'%1$s\' has been successfully returned</string>
-  <string name="reachedLoanLimit">u have reached your loan limit</string>
+  <string name="reachedLoanLimit">You have reached your loan limit</string>
 </resources>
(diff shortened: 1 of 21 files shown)


simplified-accessibility/src/main/res/values/strings.xml
```diff
@@ -8,4 +8,5 @@
   <string name="bookIsDownloading">The book \'%1$s\' has started downloading</string>
   <string name="bookIsOnHold">A reservation has been placed for the book \'%1$s\'</string>
   <string name="bookReturned">The book \'%1$s\' has been successfully returned</string>
+  <string name="reachedLoanLimit">u have reached your loan limit</string>
```
ray-lee, line 11: Typo in "u"?

ray-lee (commented):
This mostly looks good. Just a few comments below.

simplified-accessibility/src/main/res/values/strings.xml
nunommts: Actually, it was a bad copy-paste 😅 Fixed!

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogBookDetailViewModel.kt
```diff
@@ -272,6 +272,12 @@ class CatalogBookDetailViewModel(
     this.borrowViewModel.tryDelete(feedEntry.accountID, feedEntry.bookID)
   }
 
+  override fun getInitialBookStatus(feedEntry: FeedEntry.FeedEntryOPDS) {
```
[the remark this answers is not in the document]
nunommts: Yes, you're absolutely right!

ray-lee (approved):
Looks good, thanks!

PR #125, opened 2022-07-19
Added playbackRates to profile preferences

nunommts:
**What's this do?**
This PR adds a field called _playbackRates_ to the profile preferences so the user can save all of their audiobook's playback rates. When listening to an audiobook, a new entry is added to that new field with the BookID and the playback rate so it can be restored when reopening that same audiobook

There's a [ticket with a bug](https://www.notion.so/lyrasis/e72462c871a54c5d9b8b4cbf740cfe2f?v=7669bb698f0b4f7486211f75d6170b17&p=0f3ab653edd54965bc9f9280702eea47) that says the audiobook's playback rate is not saved when the user closes the book and they need to set it again when reopening the book.

_This can only be tested when [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/33) is merged_

Open the Palace app
Open any audiobook
Change the playback rate to a different value
Close the audiobook
Reopen the audiobook and verify the rate is the you previously select

This PR will only work and can only be tested after [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/33) is merged. For this to be fully completed, i.e., for the audiobooks to restore the saved rates, [this PR](https://github.com/ThePalaceProject/android-drm-audioengine/pull/9) needs to be merged too.

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-tests/src/test/resources/org/nypl/simplified/tests/books/profile-rates.json
+++ b/simplified-tests/src/test/resources/org/nypl/simplified/tests/books/profile-rates.json
@@ -12,7 +12,10 @@
       "font_family" : "READER_FONT_OPEN_DYSLEXIC",
       "color_scheme" : "SCHEME_WHITE_ON_BLACK"
     },
-    "playbackRates" : "{\"bookid1\": \"NORMAL_TIME\", \"bookid2\": \"ONE_AND_A_QUARTER_TIME\"}",
+    "playbackRates" : {
+      "bookid1" : "NORMAL_TIME",
+      "bookid2" : "ONE_AND_A_QUARTER_TIME"
+    },
     "dateOfBirth" : {
       "date" : "2006-08-07",
       "isSynthesized" : true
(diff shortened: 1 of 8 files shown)


simplified-tests/src/test/resources/org/nypl/simplified/tests/books/profile-rates.json
```diff
@@ -0,0 +1,22 @@
...
+    "hasSeenLibrarySelectionScreen" : true,
+    "useExperimentalR2" : true,
+    "showDebugSettings" : true,
+    "mostRecentAccount" : "5310437f-db1a-492e-a09f-1ceaa43303dd",
+    "readerPreferences" : {
+      "font_scale" : 100.0,
+      "font_family" : "READER_FONT_OPEN_DYSLEXIC",
+      "color_scheme" : "SCHEME_WHITE_ON_BLACK"
+    },
+    "playbackRates" : "{\"bookid1\": \"NORMAL_TIME\", \"bookid2\": \"ONE_AND_A_QUARTER_TIME\"}",
```
ray-lee, line 15: This seems strange to me. Instead of storing a string of JSON, can we make this a JSON object, like `readerPreferences` is? For example:

```
"playbackRates" : {
  "bookid1" : "NORMAL_TIME", 
  "bookid2" : "ONE_AND_A_QUARTER_TIME"
},
```

ray-lee (commented):
I just have a question about the JSON serialization.

ray-lee (approved):
Looks good, thanks!

PR #129, opened 2022-07-29
Feature/add reset password to login view

nunommts:
**What's this do?**
This PR adds a new link to handle on the authentication document regarding the user's password reset on some accounts. If that link is present on the library's authentication document, then a label is presented to the user so they can click on it and finish the process in a webview.

There's [a feature request](https://www.notion.so/lyrasis/Android-Add-password-reset-link-for-virtual-library-cards-fd28cd2c5e0d480abf217a066ae94a09?p=d25f58e32a0444469382a700c0daf7fd&pm=c) to add this to the apps so they can support this feature.

Open the app
Select "LYRASIS Reads"
Go to settings and then to the account details screen
If you're logged in, logout
Confirm [the label is there](https://user-images.githubusercontent.com/79104027/181791136-408cc740-5d09-407d-9cd7-527bd7e856c8.png)
Click on it and verify [a webview's opened so you can reset your password](https://user-images.githubusercontent.com/79104027/181791226-4df817d8-dfe0-4204-be93-01e720580297.png)

None

Yes

No

Tested by @nunommts


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountAuthenticationViewBindings.kt
```diff
@@ -181,6 +190,19 @@ sealed class AccountAuthenticationViewBindings {
...
+    override fun setResetPasswordLabelStatus(
+      status: AccountLoginButtonStatus,
+      isVisible: Boolean,
+      onClick: () -> Unit
+    ) {
+      this.resetPasswordLabel.setOnClickListener {
+        onClick()
+      }
+      this.resetPasswordLabel.isVisible = isVisible && (
+        status is AsLoginButtonEnabled || status is AsLoginButtonDisabled
```
ray-lee, line 202: Why does the visibility of the password reset link depend on the status of the login button?

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountAuthenticationViewBindings.kt
```diff
@@ -330,6 +354,18 @@ sealed class AccountAuthenticationViewBindings {
...
+    override fun setResetPasswordLabelStatus(
+      status: AccountLoginButtonStatus,
+      isVisible: Boolean,
+      onClick: () -> Unit
+    ) {
+
+      this.resetPasswordLabel.setOnClickListener {
+        onClick()
+      }
+      this.resetPasswordLabel.isVisible = isVisible && status is AsLoginButtonEnabled
```
ray-lee, line 366: Same question.

ray-lee (commented):
I just had a couple of questions below. Actually one question, two times.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountAuthenticationViewBindings.kt
nunommts: Well, the feature says that the "forgot your password" label should appear when the user is not logged in, and those two status refer to that scenario. I assumed that as long as the login button is there, regardless of its state, the label should also be there (if there's an URI too). The remaining status refer to the "Cancel" and "Logout" options, so I thought the label shouldn't be visible in these cases

ray-lee: Thanks, that makes sense. Those states are confusing.

ray-lee (approved):

PR #130, opened 2022-08-03
Fixed audiobook bookmarks not being saved after exiting the player.

nunommts:
**What's this do?**
This PR divides the bookmark saving concept into two different ways: local and remote. By doing that, we can exit the player screen and just save the bookmark remotely because now it'll be locally saved every second (implemented on [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/38)).

There's [a ticket with a bug](https://www.notion.so/lyrasis/Android-Last-reading-position-is-not-saved-after-reopening-the-audiobook-from-Biblioboard-or-Palac-b1ea2ca76af54449aa05b22f5d598f88) saying that some audiobooks are not storing the last position after exiting the player screen.

_Note: The PR can only be tested when [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/38) is merged_

Open the Palace app
Select "LYRASIS Reads"
Open the "Daykeeper" audiobook
Wait for the bookmark saving animation to happen and listen to the audiobook for about two more seconds
Return to the previous screen and back to the audiobook
Verify the UI is displaying the last position you listened to

This PR depends on [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/38)

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -190,7 +190,7 @@
         <c:change date="2022-04-29T00:00:00+00:00" summary="Fixed &quot;Unable to initialize audio engine&quot; error playing some audio books."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-07-29T15:13:53+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.11">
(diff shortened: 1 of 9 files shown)


ray-lee (approved):
Looks good to me!

PR #131, opened 2022-08-05
Fix broken state after returning a book.

nunommts:
**What's this do?**
This PR changes the way the CatalogBookViewModel recreates a "book with a status" after returning it. Instead of using the _feedEntry_ from the initial parameters, now it uses an updated feed entry that's retrieved after returning the book.

There's a [ticket with a bug](https://www.notion.so/lyrasis/There-s-an-error-trying-to-get-a-book-after-returning-it-from-My-Books-tab-370a14f9163d483885d7da4d987d40a3) explaining and showing how this bug is happening and it should be fixed so the user can get the book again directly from its details screen on the"My Books" tab instead of having to go to the catalog to retrieve it again.

You can follow the steps described in the [ticket](https://www.notion.so/lyrasis/There-s-an-error-trying-to-get-a-book-after-returning-it-from-My-Books-tab-370a14f9163d483885d7da4d987d40a3)
In the end, the button's text must be "Get" (like in the "Catalog" tab) and the book should be successfully retrieved again

No

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -190,7 +190,7 @@
         <c:change date="2022-04-29T00:00:00+00:00" summary="Fixed &quot;Unable to initialize audio engine&quot; error playing some audio books."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-08-03T10:15:04+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.11">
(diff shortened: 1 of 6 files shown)


ray-lee (approved):
Looks good to me.

PR #133, opened 2022-08-10
Added 'content description' text to back button and logo on toolbar.

nunommts:
**What's this do?**
This PR adds a content description to the back button / logo image on the toolbar. When setting the toolbar's imageview image, we verify if it's the back button or the app's logo and set its content description according to it

There's a [ticket with a bug](https://www.notion.so/lyrasis/Android-There-is-an-unlabeled-button-on-the-book-detail-view-screen-2fbc0a5aa61545e78fede8ce86dbf4aa) saying that the back button of the toolbar doesn't display a "Text" value when selecting on the inspect mode of the BrowserStack.

The [BrowserStack now displays the "Back" label](https://user-images.githubusercontent.com/79104027/183928158-0a269d52-d9f8-405e-b502-2303a4f74e2d.png) on the Inspect Mode, but this can be tested by turning the TalkBack feature of the phone in the Accessibility options, open any book on the app and then click on the "Back" button.

No

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -190,7 +190,7 @@
         <c:change date="2022-04-29T00:00:00+00:00" summary="Fixed &quot;Unable to initialize audio engine&quot; error playing some audio books."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-08-05T08:51:13+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.11">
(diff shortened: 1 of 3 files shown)


ray-lee (approved):
Easy enough!

ray-lee (changes requested):
@nunommts This looked good when I reviewed it earlier today, but now it seems to go back to older commits of the org.thepalaceproject.android.platform and bookmark_annotations/spec submodules. Could you remove those changes?

ray-lee (approved):

PR #134, opened 2022-08-10
Add ability to always show the password in the account details screen

nunommts:
**What's this do?**
This PR excludes the "Show password" checkbox from being disabled depending on the login status. It also adds a flag to determine if the password input changed after clicking on that checkbox or not, because it would cause some UI inconsistency.

There's a [ticket with a bug](https://www.notion.so/lyrasis/Android-Add-ability-to-select-Show-Password-from-the-Sign-in-Sign-out-screen-when-logged-in-553cccdd64134c0a8bf26628aa923ece) referring that when the user is logged in a library they cant' see their password unless they logout, enable the checkbox and log in again.

Open the Palace app
Go to the Account details screen
If you're not logged in, log in to that library
Verify the ["Show password" checkbox is enabled](https://user-images.githubusercontent.com/79104027/183980735-5da4a6cb-6fc5-496b-bd79-338c860fb44d.png)
Click on it and verify your password is visible

No

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -221,11 +221,11 @@
         <c:change date="2022-08-05T00:00:00+00:00" summary="Fixed a &quot;Download&quot; button appearing after returning a book, that would lead to an error if tapped."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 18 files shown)


ray-lee (approved):
Wow, it's pretty annoying that the text changed listener is called when the text didn't change.

PR #135, opened 2022-08-12
Added back button when search field appears

nunommts:
**What's this do?**
This PR sets a fixed height to the Toolbar so on the "Add Library" screen opening the SearchView doesn't cause the Toolbar's height to be improperly increased. By doing that, there was the need of removing one of the actions from the SearchView in the XML to make the UI consistent. After that, when opening the SearchView, there were two buttons: the Toolbar's back button and the SearchView clear text button. Both of them should have the same action, i.e., close the SearchView when it's open, so there was also the need of changing the Back button clicking action by checking if the toolbar has a SearchView that's not iconified, i.e., that's open.

There's a [ticket with a bug](https://www.notion.so/lyrasis/Android-The-back-arrow-disappears-when-the-search-field-appears-cc14886f53fb4154a8c4327f27491f3b) saying the back button should always visible so the user can return to the previous screen without having to close the searching field manually.

Open Palace app
Go to "Settings"
Click on "Libraries"
Click on the + icon at the top right
Click on the search icon
Verify [the searching field is open with both the back button and clear icon](https://user-images.githubusercontent.com/79104027/184376480-d047cfd3-1418-4ce5-b47a-1a3116e8a85c.png)
Click on the clear icon and verify the searching field closed
Open the searching field again
Click on the back button
Verify the searching field closed

No

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -190,7 +190,7 @@
         <c:change date="2022-04-29T00:00:00+00:00" summary="Fixed &quot;Unable to initialize audio engine&quot; error playing some audio books."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-08-05T08:51:13+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.0.11">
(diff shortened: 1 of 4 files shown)


ray-lee (approved):
Looks good.

PR #136, opened 2022-08-17
Fixed sample preview being handled as full content acquisition link

nunommts:
**What's this do?**
This PR adds a method to confirm if the relation is supported when trying to consume acquisitions. By setting the supported relations, we can prevent the sample links from being handled as full content acquisition links.

There's a [reported bug](https://www.notion.so/lyrasis/Android-apps-treat-sample-preview-links-as-full-content-acquisition-links-4e2e1de4308944ffbd6b0c363a8fb181) saying that some books are not being borrowed because of an error since the app is considering the _sample link_ to be a full acquisition link.

Open the Palace app
Choose the OverDrive Integration Test Library.
Search for *Breath of Scandal* by Sandra Brown
Open the book and click on "Get"
Confirm the book is being successfully borrowed

No

Yes

No

Tested by @nunommts


ray-lee (commented):
I'm not sure about this. I think the parser should really only be concerned with converting the XML to java. It shouldn't really have an opinion on whether or not a particular acquisition link is _supported_, because it doesn't know how that acquisition link is going to be used.

For example, the preview (sample) acquisition type should not be supported by the full book reader, but it should be supported by the preview viewer, when we implement that. This would make that acquisition link unavailable in both cases, but we only want it not to be used in one case.

What do you think about moving this check to a higher level?

nunommts:
@ray-lee Yes, you're right. I agree with you, so I'll be moving these changes to somewhere else on a higher level. However, the title the bug refers to seems to have had its feed entry changed and the bug is no longer happening. It seems the sample link has changed from _http://opds-spec.org/acquisition/sample_ to _preview_

simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/BorrowAcquisitions.kt
```diff
@@ -20,7 +21,9 @@ object BorrowAcquisitions {
     entry: OPDSAcquisitionFeedEntry
   ): OPDSAcquisitionPath? {
     val paths = OPDSAcquisitionPaths.linearize(entry)
-    val filtered = paths.filter { support.isSupportedPath(it.asMIMETypes()) }
-    return filtered.firstOrNull() ?: return null
+    return paths
+      .filter { support.isSupportedPath(it.asMIMETypes()) }
+      // when borrowing a book, we need to ignore if the acquisition is for a sample/preview
+      .firstOrNull { it.source.relation != OPDSAcquisition.Relation.ACQUISITION_SAMPLE }
```
ray-lee, line 27: Shouldn't this also exclude "preview"?

simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAcquisitionFeedEntryParser.java
```diff
@@ -284,7 +284,7 @@ private void tryConsumeAcquisitions(
     final String rel_text)
     throws OPDSParseException {
 
-    if (rel_text.startsWith(ACQUISITION_URI_PREFIX_TEXT)) {
+    if (rel_text.startsWith(ACQUISITION_URI_PREFIX_TEXT) || rel_text.toLowerCase().equals("preview")) {
```
ray-lee, line 287: "preview" should probably be in a constant, because we'll probably need to use it again in other places.

ray-lee (commented):
I think there's a bug in the BorrowAcquisitions change (see comment below). I would feel better if there are some unit tests to make sure it's working, and to prevent any regression in the future.

simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAcquisitionFeedEntryParser.java
nunommts: This shouldn't be here, sorry! I added this in order to reproduce the error with the latest changes from the server and it ended up here by mistake.

simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/BorrowAcquisitions.kt
nunommts: I didn't add the "preview" because we still haven't moved on with the samples feature. I replied to your other comment, but basically the "preview" code was pushed by mistake.

ray-lee: Ah, that makes sense.

simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAcquisitionFeedEntryParser.java
ray-lee: Ok, thanks.

ray-lee (approved):
Looks good.

PR #137, opened 2022-08-23
Switched default PDF reader to the PDF-JS Reader

nunommts:
**What's this do?**
This PR sets the PDF-JS reader as the default PDF reader and updates the dev settings so the user can set the old PDF reader if they want

The PDF-JS reader seems to be stable and working very well so there's a [feature request](https://www.notion.so/lyrasis/Android-Promote-new-PDF-reader-to-be-the-default-reader-912b0a24b12042fcbf444f1c237902ec) to promote it as the default reader.

Open the Palace app
Choose "LYRASIS Reads" library
Open a pdf book from _New Biblioboard Test_ lane
Confirm the pdf reader [is the PDF-JS reader](https://user-images.githubusercontent.com/79104027/186181938-3e88af4b-0825-47d8-8b08-f79b3fedf283.png)
Go to settings
Enable debug options
Enable "Set old PDF reader" option
Return to the PDF book
Confirm the pdf reader [is the old PDF reader](https://user-images.githubusercontent.com/79104027/186181966-6ae86b95-299e-474d-8c08-c77586eb4cce.png)

No

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
+++ b/simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
@@ -12,8 +12,9 @@ import org.nypl.simplified.viewer.spi.ViewerPreferences
 class PdfViewerProviderTest {
 
   @Test
-  fun supportsPdfBooksWhenEnabled() {
+  fun doesNotSupportPdfBooksWhenDisabled() {
     val preferences = ViewerPreferences(
+      // by setting the old PDF reader setting to true, we're disabling the PDF-JS reader
       flags = mapOf(
         "enableOldPDFReader" to true
       )
@@ -27,8 +28,9 @@ class PdfViewerProviderTest {
   }
 
   @Test
-  fun doesNotSupportPdfBooksWhenDisabled() {
+  fun supportsPdfBooksWhenEnabled() {
     val preferences = ViewerPreferences(
+      // by setting the old PDF reader setting to false, we're enabling the PDF-JS reader
       flags = mapOf(
         "enableOldPDFReader" to false
       )
--- a/simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
+++ b/simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
@@ -15,22 +15,7 @@ class PdfViewerProviderTest {
   fun supportsPdfBooksWhenEnabled() {
     val preferences = ViewerPreferences(
       flags = mapOf(
-        "enablePDFJSReader" to true
-      )
-    )
-
-    val book = Mockito.mock(Book::class.java)
-    val format = Mockito.mock(BookFormat.BookFormatPDF::class.java)
-    val provider = PdfViewerProvider()
-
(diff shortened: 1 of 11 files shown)


simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
```diff
@@ -15,30 +15,30 @@ class PdfViewerProviderTest {
   fun supportsPdfBooksWhenEnabled() {
```
ray-lee, line 15: This is now asserting the opposite, so the name should change to `doesNotSupportPdfBooksWhenDisabled`.

simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
```diff
@@ -15,30 +15,30 @@ class PdfViewerProviderTest {
...
     val book = Mockito.mock(Book::class.java)
     val format = Mockito.mock(BookFormat.BookFormatPDF::class.java)
     val provider = PdfViewerProvider()
 
-    Assertions.assertTrue(provider.canSupport(preferences, book, format))
+    Assertions.assertFalse(provider.canSupport(preferences, book, format))
   }
 
   @Test
   fun doesNotSupportPdfBooksWhenDisabled() {
```
ray-lee, line 30: This is now asserting the opposite, so the name should change to `supportsPdfBooksWhenEnabled`.

simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
```diff
@@ -15,30 +15,30 @@ class PdfViewerProviderTest {
   fun supportsPdfBooksWhenEnabled() {
     val preferences = ViewerPreferences(
       flags = mapOf(
-        "enablePDFJSReader" to true
+        "enableOldPDFReader" to true
```
ray-lee, line 18: I think we need a comment here to explain that setting `enableOldPDFReader`to true _disables_ the _current_ (pdf.js) reader. Otherwise it will be confusing.

simplified-tests/src/test/java/org/nypl/simplified/tests/pdf/PdfViewerProviderTest.kt
```diff
@@ -15,30 +15,30 @@ class PdfViewerProviderTest {
...
-    Assertions.assertTrue(provider.canSupport(preferences, book, format))
+    Assertions.assertFalse(provider.canSupport(preferences, book, format))
   }
 
   @Test
   fun doesNotSupportPdfBooksWhenDisabled() {
     val preferences = ViewerPreferences(
       flags = mapOf(
-        "enablePDFJSReader" to false
+        "enableOldPDFReader" to false
```
ray-lee, line 33: I think we need a comment here to explain that setting `enableOldPDFReader` to false _enables_ the _current_ (pdf.js) reader. Otherwise it will be confusing.

ray-lee (commented):
This mostly looks good. I made a few comments about the changes to the test.

ray-lee (approved):
Looks good.

PR #139, opened 2022-09-01
Fixed back button action on some screens

nunommts:
**What's this do?**
This PR changes the way the back button actions are handled. When the user goes to the account selection from the catalog, we're now overriding the logo click action so the app can perform the action of returning to the catalog. When the user is on the onboarding screen and wants to choose a library, the account list screen is now being removed from the fragment manager so the user can return to the previous screen. Also, this PR changes the source tab that called the account list screen as it can be opened from two different areas: the Catalog and the Settings.

There's a [reported bug](https://www.notion.so/lyrasis/Android-The-Back-button-doesn-t-return-to-the-Catalog-screen-a21a148239e34935a70794bf697644c9) saying the back button is not returning the user to the Catalog when opening the account list screen from there. There's also a [reported bug](https://www.notion.so/lyrasis/Android-The-Back-button-doesn-t-return-to-the-previous-screen-16718360278c45079668079619fa0939) saying the back button is not working on the onboarding screen

_First test_
Uninstall the app and install it again.
Open the palace app
Skip the Onboarding
Click on "Find Library"
Click on the back button at the top left corner
Verify the app returned to the previous screen

_Second Test_
Open the Palace app
On the Catalog, click on the logo at the top left corner
Click on "Add Library"
Press the back button at the top left corner
Verify the app returned to the Catalog

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -221,7 +221,7 @@
         <c:change date="2022-08-05T00:00:00+00:00" summary="Fixed a &quot;Download&quot; button appearing after returning a book, that would lead to an error if tapped."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-08-31T00:16:57+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.2.0">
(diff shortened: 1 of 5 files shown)


ray-lee (approved):
Looks good to me.

PR #140, opened 2022-09-01
Updated method to calculate the width of the SearchView

nunommts:
**What's this do?**
This PR updates the logic of how the maximum width for the SearchView on the account list screen is calculated. Instead of relying on the number of icons, it's now using the start inset navigation value of the toolbar.

Even though [this bug](https://www.notion.so/lyrasis/Android-The-back-arrow-disappears-when-the-search-field-appears-cc14886f53fb4154a8c4327f27491f3b) was fixed, the solution could be improved

_Note: I've ran the app on [the bug reported devices](https://www.notion.so/lyrasis/Android-The-back-arrow-disappears-when-the-search-field-appears-cc14886f53fb4154a8c4327f27491f3b) and the results were:
[Samsung Galaxy S9-8](https://user-images.githubusercontent.com/79104027/187974515-64f57071-f32f-41a0-a840-40871513b9e8.png)
[XiaoMI Mi9](https://user-images.githubusercontent.com/79104027/187974520-eb12dc02-e518-444d-94ec-14f2c0f2b108.png)
[Samsung Galaxy S22-12](https://user-images.githubusercontent.com/79104027/187974521-d5277c7d-b800-4fc1-85b0-364e40f1661d.png)
[Samsung Galaxy S22 Ultra-12](https://user-images.githubusercontent.com/79104027/187974524-2185ec87-adce-4856-a1bf-b87805528547.png)
_

Open the app
Go to Settings
Click on "Libraries"
Click on the "+" at the top right corner
Click on the search icon
Verify the SearchView is not overlapping the back button

None

No

No

Tested by @nunommts

### DIFF
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryFragment.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryFragment.kt
@@ -1,7 +1,6 @@
 package org.nypl.simplified.ui.accounts
 
 import android.annotation.SuppressLint
-import android.app.Activity
 import android.content.Context
 import android.location.LocationManager
 import android.os.Bundle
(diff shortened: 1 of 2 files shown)


ray-lee (approved):
Looks good to me.

ray-lee:
@nunommts This needs a conflict resolved.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryFragment.kt
```diff
@@ -43,6 +44,10 @@ class AccountListRegistryFragment : Fragment(R.layout.account_list_registry) {
   private val subscriptions = CompositeDisposable()
   private val listener: FragmentListenerType<AccountListRegistryEvent> by fragmentListeners()
 
+  private val requestLocationPermission = registerForActivityResult(
+    ActivityResultContracts.RequestPermission(),
+    ::getLocation
+  )
```
ray-lee, line 50: Is this intended to be part of this change? It seems to have come in with the last merge, but I'm not sure how it's related.

nunommts: Yes, it is. I deleted that because it was being unused, but there are some commented code that is using that and that's why it was considered to be unused. However, I thought it would be better to restore it in case the code gets uncommented and in that way we can save some time.

PR #141, opened 2022-09-01
Removed delete option from book details screen

nunommts:
**What's this do?**
This PR removes the 'delete' option from a book details screen.

There's a [feature request](https://www.notion.so/lyrasis/Android-There-is-Return-button-instead-of-Delete-in-a-book-from-Palace-Bookshelf-collection-in-LYRA-1a1060dc48054d8ea2cf67e558431f52) to remove this option so the user doesn't get confused with the available options.

Open the Palace app.
Select LYRASIS Reads
Search for "The Prince and the Pauper"
Open the book and click on "Get"
Verify [there are only two available options](https://user-images.githubusercontent.com/79104027/188028723-1cde3f0e-6d0b-4ed3-a637-561fb9d287cb.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogPagedViewHolder.kt
+++ b/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogPagedViewHolder.kt
@@ -22,6 +22,14 @@ import org.nypl.simplified.feeds.api.FeedEntry
 import org.nypl.simplified.feeds.api.FeedEntry.FeedEntryCorrupt
 import org.nypl.simplified.feeds.api.FeedEntry.FeedEntryOPDS
 import org.nypl.simplified.futures.FluentFutureExtensions.map
+import org.nypl.simplified.opds.core.OPDSAvailabilityHeld
+import org.nypl.simplified.opds.core.OPDSAvailabilityHeldReady
+import org.nypl.simplified.opds.core.OPDSAvailabilityHoldable
+import org.nypl.simplified.opds.core.OPDSAvailabilityLoanable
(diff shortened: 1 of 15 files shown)


ray-lee (changes requested):
This is more complicated than just removing the Delete button. In cases where we currently show both Return and Delete buttons, it's probably fine to just remove the Delete. But there are cases where we only show a Delete button, like in the Palace Bookshelf library. In that case we would need to keep the functionality, but change the button label to "Return". I'm not sure about all of the cases where there might only be a Delete button, so part of this is to make sure we get them all.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogPagedViewHolder.kt
```diff
@@ -508,16 +508,6 @@ class CatalogPagedViewHolder(
...
-    } else if (isBookDeletable(book)) {
-      this.idleButtons.addView(this.buttonCreator.createButtonSpace())
-      this.idleButtons.addView(
-        this.buttonCreator.createDeleteButton(
-          onClick = {
-            this.listener.delete(this.feedEntry as FeedEntryOPDS)
-          },
-          heightMatchParent = true
-        )
-      )
```
ray-lee, line 520: Can you explain why this was deleted, Instead of keeping it, but labeling it "Return" like you did in other places? I'm not saying it's wrong, I'm just not sure about it.

nunommts: I've missed it! It was supposed to be restored as well. Thanks for pointing that out.

ray-lee (approved):
I think this looks good, thanks!

PR #142, opened 2022-09-21
Fixed logging out not working sometimes

nunommts:
**What's this do?**
This PR updates the login state of the account as soon as the user presses the "Log out" button instead of waiting for the logging out task to actually start. When the user logs in, there are some background operations that may take some time to be completed, so if the user immediately (or not) logs out, the logout button will be disabled as expected but nothing seems to happen after that. If the user waits long enough, the logging out will actually start because the pending operations were finished. This PR doesn't stop those background tasks from happening but updates the UI so the user is aware that something is happening and they need to wait until they are completely logged out

There's a [bug that was reported](https://www.notion.so/lyrasis/Android-Investigate-the-cases-of-Log-in-button-unavailability-f3dfc83441c54ba79618a2c5c08e7e85) as the login button being unavailable after pressing the "Log out" button. Nothing appeared to happen, so the user would be in a "dead" state where they couldn't reinsert the log in credentials but they weren't actually logged out. Even though the logging out action would still happen after some time (as it can be seen [here](https://user-images.githubusercontent.com/79104027/191543008-05889a13-df3b-464e-8363-4626400df0a1.mp4)), there was the need to change this behavior so the user doesn't get lost.

Open the Palace app
Select the "Lyrasis Reads" library
Go to the account details and Log in (if you're logged in, log out first)
Immediately after logging in, press "Log out"
Verify the [loading is appearing while the remaining log in background tasks are happening and that after some time the logout task starts](https://user-images.githubusercontent.com/79104027/191542812-85de9ec8-ac9b-4c4c-8e75-13f70f249d56.mp4)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -221,7 +221,7 @@
         <c:change date="2022-08-05T00:00:00+00:00" summary="Fixed a &quot;Download&quot; button appearing after returning a book, that would lead to an error if tapped."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-09-19T04:00:37+00:00" is-open="false" ticket-system="org.nypl.jira" version="1.2.0">
+    <c:release date="2022-09-19T00:00:00+00:00" is-open="false" ticket-system="org.nypl.jira" version="1.2.0">
       <c:changes>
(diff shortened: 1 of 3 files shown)


ray-lee (approved):
Looks good to me.

PR #143, opened 2022-09-22
Fixed app crashing when opening TOC or settings before loading the PDF

nunommts:
**What's this do?**
This PR checks if the PDF Viewer WebView has been initialized before applying any of the selected options by the user. When the PDF is loading, the user may already select the Settings or TOC options and if the WebView hasn't been initialized yet nothing should happen.

There's a [reported bug](https://www.notion.so/lyrasis/Android-PDF-There-is-a-redirection-to-the-Catalog-or-the-app-stops-when-we-tap-TOC-or-Settings-str-f441d733c81e4486b8425d000652c6ee) saying the app is crashing when selecting the TOC/Settings option while the PDF is being loaded.

Open the Palace app
Select "LYRASIS Reads"
Search for "Who lives in a tree"
Download the PDF and open it
Select both TOC and Settings option before the PDF loads
Verify nothing happens and the app doesn't crash

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -237,9 +237,10 @@
         <c:change date="2022-09-19T00:00:00+00:00" summary="Fixed a crash that occurred on some Samsung devices."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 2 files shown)


ray-lee (approved):
Looks good!

PR #144, opened 2022-09-27
Added content description to back button on reader screen

nunommts:
**What's this do?**
This PR adds a content description to the back button on the toolbar of the reader screen.

There's a [ticket with a bug](https://www.notion.so/lyrasis/Android-There-are-unlabeled-buttons-on-the-book-reading-and-listening-screens-1847df33398945e487c8f0e6b3a136df) saying that the back button of the reader's toolbar doesn't display a "Text" value when selecting on the inspect mode of the BrowserStack.

The [BrowserStack now displays the "Back" label](https://user-images.githubusercontent.com/79104027/192505454-7dd41e87-dcdb-4fd3-9846-449329c60a01.png) on the Inspect Mode, but this can be tested by:
1. Turn the TalkBack feature of the phone in the Accessibility options
2. Open the Palace app
3. Open any book to start reading it
4. Click on the "Back" button.

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -237,11 +237,12 @@
         <c:change date="2022-09-19T00:00:00+00:00" summary="Fixed a crash that occurred on some Samsung devices."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-09-22T17:25:39+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.3.0">
+    <c:release date="2022-09-27T10:36:01+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.3.0">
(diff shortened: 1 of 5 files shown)


ray-lee (commented):
@nunommts This only makes a change for pdfs, right? Do we need to do anything for epubs and audiobooks, or do they work already?

nunommts:
@ray-lee That's a good point. I only did this for the PDFs because it was reported on the ticket. I checked the code and if there's anything to do regarding this for the epubs and audiobooks it should be done on their respective modules, so I'll create the PRs for those too

ray-lee (approved):

PR #146, opened 2022-10-08
Fixed login button being disabled when reentering the app

nunommts:
**What's this do?**
This PR removes two lines of code that had impact on how the UI behaves in the account detail screen when the user send the app to background. There's no point in clearing the views list and setting the image to null and when the user returns those two objects weren't restored, which was causing the login button to being disabled and the account logo image to disappear.

There's a comment in [this ticket](https://www.notion.so/lyrasis/Android-Investigate-the-cases-of-Log-in-button-unavailability-f3dfc83441c54ba79618a2c5c08e7e85) saying the login button is disabled when reentering the app

Open the Palace app
Select a library where you're logged in to
Logout
Wait for the logout to be completed and send the app to background
Return to the app and insert your barcode and pin
Confirm the "Login" button is enabled

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -237,7 +237,7 @@
         <c:change date="2022-09-19T00:00:00+00:00" summary="Fixed a crash that occurred on some Samsung devices."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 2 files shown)


ray-lee (approved):
Looks good to me.

PR #147, opened 2022-10-12
Fixed sidebar automatically opening when opening some pdfs

nunommts:
**What's this do?**
This PR adds a listener to the webview's page loading to be aware of when the pdf is loaded so it can check if the sidebar is open and, if so, close it.

There's a [ticket with a bug](https://www.notion.so/lyrasis/Android-Pdfs-in-Palace-Bookshelf-library-are-opened-with-already-opened-TOC-page-8a3b6a90a004490ebc083c6ffbddf2a2) reporting this unwanted behavior that is causing a bad UX.

Open the Palace app
Selected "Palace Bookshelf" library
Get a pdf (eg. “Deep into Pharo”)
Tap “Read”
Confirm [the sidebar is not automatically opening when the pdf is loaded](https://user-images.githubusercontent.com/79104027/195463662-9ff8a219-9fd0-4013-b58b-e48975b90e8f.mp4)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -237,7 +237,7 @@
         <c:change date="2022-09-19T00:00:00+00:00" summary="Fixed a crash that occurred on some Samsung devices."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-10-08T21:06:00+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.3.0">
+    <c:release date="2022-10-12T22:59:39+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.3.0">
       <c:changes>
         <c:change date="2022-09-21T00:00:00+00:00" summary="Removed delete option from book details screen."/>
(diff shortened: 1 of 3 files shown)


nunommts:
@ray-lee I didn't know that, but I've tried to add it to the url (_http://localhost:${it.port}/assets/pdf-viewer/viewer.html?pagemode=none&file=%2Fbook.pdf#page=${this.documentPageIndex}_) and the bug still happens

ray-lee:
That makes me think that the PDF itself is asking for the navigation to be open. In which case, this may not be a bug.

ray-lee:
I verified that the book in the Notion ticket does in fact request that the outline be open. The PDF reader is respecting this instruction. (See https://github.com/mozilla/pdf.js/issues/8972).

As discussed in today's tech team meeting, this isn't actually a bug, so I'm closing this PR. If we revisit this later and decide that we want to ignore the instruction from the PDF, the above link contains instructions.

PR #148, opened 2022-10-14
Externalize LCP build configuration.

ray-lee:
**What's this do?**

This is a refinement of #145, to avoid breaking PR builds.

This moves the configuration for downloading the LCP decryption library into an external script. It also removes dependencies on the LCP library from modules where it wasn't being used, so that the only remaining use is in `simplified-app-palace`.

This makes the LCP build recipe private. Notion: https://www.notion.so/lyrasis/Move-LCP-recipe-to-properties-c508747ff48b416a9a6f31096e1f8e03

Verify that LCP books (e.g. epubs from the DPLA Titles lane in LYRASIS Reads, such as "The Great Illyrian Revolt" can be downloaded.

https://github.com/ThePalaceProject/mobile-certificates/pull/23 should be merged first.

No, this is not an externally visible change.

The README has been updated.

@ray-lee ran some builds and checked LCP downloads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,14 @@ audiobook player module, but there is also an NYPL-developed Findaway
 module for Findaway licensees. Please get in touch with us if you have
 a Findaway license and want to produce a Findaway-enabled build.
 
+#### LCP DRM Support
+
+The project uses Readium's liblcp module to provide support for LCP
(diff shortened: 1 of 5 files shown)


nunommts (approved):
Looks good!

PR #153, opened 2022-12-02
Feature/hide toolbar on pdf reader

nunommts:
**What's this do?**
This PR adds logic to hide/show the PDF Reader toolbar when pressing the PDF. This also adds a TextView with the PDF's title so it can be fully displayed when the toolbar is hidden.

There's a [ticket](https://www.notion.so/lyrasis/Android-PDF-Hide-the-toolbar-to-make-the-whole-book-title-visible-784c98e0b54748bab414a5ee4fcd329e) requesting to add the ability of hiding the toolbar in the PDF viewer so the user can read the entire pdf title without going back to the book details screen

Open the Palace app
Open any PDF (e.g., _Working on the Pony Express_ on LYRASIS Reads)
Click on the PDF
Confirm the [toolbar is hidden and the PDF title is fully displayed](https://user-images.githubusercontent.com/79104027/205403736-096f32d6-7a22-48bd-ae96-f1356d896b03.png)
Click on the PDF again
Confirm the toolbar is visible again

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-viewer-pdf-pdfjs/src/main/assets/pdf-viewer/viewer.js
+++ b/simplified-viewer-pdf-pdfjs/src/main/assets/pdf-viewer/viewer.js
@@ -11,7 +11,22 @@ PDFViewerApplication.initializedPromise.then(() => {
 
     PDFListener.onPageChanged(PDFViewerApplication.page);
   }, { external: true });
+
+  // Detect click events on the PDF viewer.
+
+  const viewers = document.getElementsByClassName("pdfViewer");
+
+  if (viewers.length > 0) {
+    viewers[0].addEventListener("click", onPDFViewerClick, { passive: true });
+  }
 });
+
+/**
+ * Notify the activity of a click on the PDF viewer.
+ **/
+function onPDFViewerClick() {
+  PDFListener.onPageClick();
+}
 
 /**
  * Toggle the sidebar.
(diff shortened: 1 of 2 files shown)


ray-lee (commented):
This doesn't seem to work for me (Pixel 5, Android 13). I can tap a lot of times on the page, and nothing happens. It does eventually toggle the toolbar, but only after 20-30 taps. I'm not sure what I'm doing differently when it works.

ray-lee (commented):
Tapping to toggle the toolbar works now, but now I can't drag to scroll the page. The page doesn't move at all.

ray-lee (commented):
Thanks, scrolling works now.

I only now noticed that when you tap on the - and + buttons or the page number input in the bottom toolbar, the top toolbar toggles. This seems weird to me. I would only expect it to toggle when you tap on the page itself.

I'm starting to think it might be better to detect the tap using JavaScript inside the WebView, where it can determine if the tap is on the page, and fire an event (or call a callback) into Java. Does that make sense? I can look into it if you're not comfortable with the JavaScript.

nunommts:
@ray-lee Actually, that was one of my approaches. I tried adding it via Javascript but I wasn't successfull with that as I don't know that much of JS. So, if you don't mind, you can try doing that because it seems like the best and cleaneste approach.

ray-lee:
@nunommts I rebased this branch, and moved the click detection code into JavaScript (see last commit). Please review when you get a chance.

nunommts:
@ray-lee I've reviewed and tested and it looks good! I can't approve this PR because I created it, but feel free to merge this whenever you can.

PR #156, opened 2022-12-15
Added "About Palace" option to Settings screen

nunommts:
**What's this do?**
This PR adds the "About Palace" option to the Settings screen and changes the way how a preference is opened. If the document configuration's name is not null or blank, then the behavior remains the same, but if it's name or file are null, then the returned URI is the _remoteURI_ so it can be opened and handled by the WebView instead of opening a local file.

There's a [feature request](https://www.notion.so/lyrasis/Android-There-is-no-an-About-Palace-item-0d3449f757bc4be2a7d08017f58223e8) to include this option on the settings of the Android app as it's already on the iOS settings screen.

Open the Palace app
Go to Settings
Confirm the ["About Palace" option](https://user-images.githubusercontent.com/79104027/207855237-cc773af7-2946-4679-994a-ad2ee20f41e0.png) is there
Click on "About Palace"
Confirm the ["About Palace" contents are displayed](https://user-images.githubusercontent.com/79104027/207855601-8bdcb59c-1406-48ee-bf22-845e295f2fbc.png)

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -256,10 +256,11 @@
         <c:change date="2022-11-29T00:00:00+00:00" summary="Fixed audiobooks not pausing when headphones are unplugged."/>
       </c:changes>
     </c:release>
-    <c:release date="2022-12-14T14:42:56+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.4.0">
+    <c:release date="2022-12-15T11:43:00+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.4.0">
       <c:changes>
         <c:change date="2022-12-12T00:00:00+00:00" summary="Changed the Switch's enabled color to green."/>
-        <c:change date="2022-12-14T14:42:56+00:00" summary="Downgraded Gradle version."/>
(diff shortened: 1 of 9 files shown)


ray-lee (commented):
This looks fine. I noticed that there aren't any unit tests at all for the simplified-documents module. Could you add some? We should have tests that verify that the right thing happens for each combination of `name` and `remoteURI` being null/not null in the DocumentConfiguration.

We should also add documentation in Notion about how document links work in the app. Adding documentation is supposed to be a requirement for closing tickets, and I'd like us to get better at it. We don't have a section yet for Android feature documentation, but I'll create one, and add documentation for this as an example going forward.

nunommts:
@ray-lee Sure, I can do that!

ray-lee (approved):
Looks good. Thanks for adding the tests.

ray-lee:
@nunommts Actually, it looks like one of the tests is failing. Could you take a look?

nunommts:
@ray-lee Sure, I've been trying to fix that failing test in the past few working days. It seems an issue related with the AssetManager not being mocked.

ray-lee (approved):
Looks good.

PR #160, opened 2023-01-19
Updated audiobook sleep timer stored value so the can be resumed

nunommts:
**What's this do?**
This PR updates the existing logic to save/restore the current audio book sleep timer value but it now updates the value every time the timer changes so the remaining duration can be stored and the timer doesn't get restarted every time the audiobook is opened.

There's a comment in [this ticket](https://www.notion.so/lyrasis/Audiobook-sleep-timer-not-saved-after-closing-book-bab815834a374893a0588ca96fc8caa7) saying the timer shouldn't be restarted when reopening the audiobook but resumed.

Follow [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/55) testing steps once that and [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/56) are merged

[This PR](https://github.com/ThePalaceProject/android-audiobook/pull/55) and [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/56) need to be merged first

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-profiles/src/main/java/org/nypl/simplified/profiles/ProfileDescriptionJSON.kt
+++ b/simplified-profiles/src/main/java/org/nypl/simplified/profiles/ProfileDescriptionJSON.kt
@@ -12,7 +12,6 @@ import org.joda.time.DateTime
 import org.joda.time.format.DateTimeFormatter
 import org.joda.time.format.DateTimeFormatterBuilder
 import org.librarysimplified.audiobook.api.PlayerPlaybackRate
-import org.librarysimplified.audiobook.api.PlayerSleepTimerConfiguration
 import org.nypl.simplified.accounts.api.AccountID
 import org.nypl.simplified.files.FileUtilities
 import org.nypl.simplified.json.core.JSONParseException
@@ -394,7 +393,7 @@ object ProfileDescriptionJSON {
   private fun deserializeSleepTimers(
     objectMapper: ObjectMapper,
     node: ObjectNode?
-  ): Map<String, PlayerSleepTimerConfiguration> {
+  ): Map<String, Long> {
 
     val str = JSONParserUtilities.getObjectOrNull(
       node, "sleepTimers"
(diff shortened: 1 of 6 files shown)


ray-lee (commented):
@nunommts For some reason the checks didn't run on this PR. Could you push a commit, to get them to run? There's a conflict in the changelog that needs to be resolved anyway.

Also, does this work when the "End of file" sleep timer is selected? I'm curious, since that isn't a specific amount of time.

nunommts:
@ray-lee Actually, that wouldn't work because there's more than a way to handle the "null" values. I'm updating it right now, but I'll need to create a new PR for the audiobook submodule that needs to be merged first.

ray-lee (commented):
@nunommts The checks are failing because there's a lint error.

nunommts:
@ray-lee Fixed!

simplified-profiles/src/main/java/org/nypl/simplified/profiles/ProfileDescriptionJSON.kt
```diff
@@ -408,7 +407,11 @@ object ProfileDescriptionJSON {
     )
 
     return map.mapValues { entry ->
-      PlayerSleepTimerConfiguration.valueOf(entry.value)
+      try {
+        entry.value.toLong()
+      } catch (exception: NumberFormatException) {
+        null
```
ray-lee, line 413: Does this mean that if there's an unexpected/non-numeric value in the json, the sleep timer will be set to end-of-file? I think it should be set to be off if there is an unexpected value. You may need to check for null explicitly before calling `toLong`, in order to check for end-of-file.

Also, from the test, it looks like older versions of the app were saving values like "MINUTES_15" and "MINUTES_30", right? If so, I think we need to check for these as well, for backward compatibility.

ray-lee: @nunommts Yes, you're right. If restoring the sleep timers wasn't working before, then I don't think we need to worry about handling those old values.

ray-lee (approved):
Looks good!

PR #166, opened 2023-02-15
Added dialog to allow the user to move to the last saved reading page or stay at the current one

nunommts:
**What's this do?**
This PR adds logic to verify if there's a newer last reading position retrieved from the server on EPUB and PDF readers. If so, then a prompt is displayed to the user so they can decide if they want to move to that saved position or remain at the current one.

There are two features, [one for the EPUB reader](https://www.notion.so/lyrasis/Android-Add-prompt-to-move-to-newer-epub-reading-position-or-stay-at-current-position-c3b50ba5a0c5453bb22c532d5dd48a64) and [other for the PDF reader](https://www.notion.so/lyrasis/Android-Add-prompt-to-move-to-newer-PDF-reading-position-or-stay-at-current-position-76c5d5a5712b4a4b9cefb07464653705), to be added to the Android version of the Palace app

_Note: You'll need to devices to properly test this_

Open the Palace app on both devices
Select any library (the same on both devices)
Download the same Ebook on both devices
Open the Ebook on device A, swipe a few pages and go back
Open the Ebook on device B and confirm you're automatically redirected to the page you were on device A (the behavior remains the same in this case)
Still on device B, swipe a few pages and go back
Reopen the Ebook on device A and confirm a prompt is shown
Click on "Move" and confirm you're redirected to the page you were on Device B
Still on device A, swipe a few pages and go back
Reopen the Ebook on device B and confirm a prompt is shown
Click on "Stay" and confirm you stay on the previously saved page on device B
Manually scroll to the page you were on device A and go back
Reopen the Ebook on device A and confirm the prompt is not shown because the pages are the same

Repeat the above steps but on a PDF book

None

Yes

No

Tested by @nunommts


ray-lee (commented):
@nunommts You may want to double check with David, but if I'm reading the ticket correctly, the dialog should only be displayed when the saved bookmark on the server is newer than the locally saved reading position. So I would only expect to see it if I read a book on a second device, leave off at a different position, and open the book again on the first device.

nunommts:
@ray-lee Yes, you're right. I've clearly misunderstood the ticket. I'll update this PR

ray-lee (approved):
This looks good. We should probably think about retiring the old pdf reader so that you don't need to keep updating it in the future.

nunommts:
@ray-lee Do you want me to do that on this PR or can I merge this one and remove the old pdf reader in a near future?

ray-lee:
@nunommts I meant for the future. This is good to merge.

PR #169, opened 2023-02-23
Change position of 'Remove' button

nunommts:
**What's this do?**
This PR updates the button creation logic to place the "Remove" after the "Get" button on a book reservation item.

There's a [feature request](https://www.notion.so/lyrasis/Android-Get-button-doesn-t-appear-first-in-the-Reservations-tab-92d7261bfe974a2ea42aea527b7e616b) saying the iOS app has the buttons in a different order than the Android app and they should have both the same order.

Open the Palace app
If you already have a book reservation that is available, go to the "Reservations" tab and confirm the "Get" button appears before the "Remove" button
If you have no reservations at all, reserve a book and wait until it's available.

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -268,7 +268,7 @@
         <c:change date="2023-01-27T00:00:00+00:00" summary="Updated audiobook sleep timer stored value so the can be resumed instead of recreated when reopening an audiobook."/>
         <c:change date="2023-02-13T00:00:00+00:00" summary="Added support for book previews."/>
         <c:change date="2023-02-20T00:00:00+00:00" summary="Fixed some audio books not appearing in the bookshelf when offline."/>
-        <c:change date="2023-02-23T19:27:04+00:00" summary="Change position of 'Remove' button."/>
+        <c:change date="2023-02-23T19:27:04+00:00" summary="Moved 'Remove' button after 'Get' button on Reservations screen."/>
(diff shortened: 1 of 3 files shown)


README-CHANGES.xml
```diff
@@ -267,7 +267,8 @@
         <c:change date="2023-01-18T00:00:00+00:00" summary="Display the search query on the search bar."/>
         <c:change date="2023-01-27T00:00:00+00:00" summary="Updated audiobook sleep timer stored value so the can be resumed instead of recreated when reopening an audiobook."/>
         <c:change date="2023-02-13T00:00:00+00:00" summary="Added support for book previews."/>
-        <c:change date="2023-02-20T20:13:35+00:00" summary="Fixed some audio books not appearing in the bookshelf when offline."/>
+        <c:change date="2023-02-20T00:00:00+00:00" summary="Fixed some audio books not appearing in the bookshelf when offline."/>
+        <c:change date="2023-02-23T19:27:04+00:00" summary="Change position of 'Remove' button."/>
```
ray-lee, line 271: This is probably too vague for an end user to know what to look for. I would prefer something like "Moved 'Remove' button after 'Get' button on Reservations screen."

ray-lee (commented):
The change log should be a bit more descriptive. Otherwise, this looks good.

PR #171, opened 2023-03-15
Upgrade actions to use node 16.

ray-lee:
**What's this do?**

This upgrades github actions that are used in our workflows to versions that use Node 16 instead of Node 12.

GitHub is disabling actions that use Node 12, which causes this warning when we run actions:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-java@v1, actions/upload-artifact@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

CI builds should continue to run successfully. The Node.js 12 warning should no longer appear.

None.

No, this is not a user visible change.

n/a

The PR build works, but we need to merge this to main to be able to run some of the other actions.

### DIFF
--- a/.github/workflows/android-main.yml
+++ b/.github/workflows/android-main.yml
@@ -9,12 +9,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Checkout submodules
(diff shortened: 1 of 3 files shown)


nunommts (approved):
Looks good!

PR #173, opened 2023-03-24
Update reading position on play, pause, and stop

jdempcy:
**What's this do?**
Save listening position on play, pause, and stop

Reduce server load
https://www.notion.so/lyrasis/Android-Save-listening-position-less-frequently-ff7c6dc43caa492d98105a48e56d4c26

Open an audiobook, navigate to another chapter in the book. Exit the audiobook and open the same audiobook within the same account on a different device and confirm the bookmark was saved.

You can also confirm that locations within chapters are saved and synced between different devices, in the same book and same account. 

Related to: https://github.com/ThePalaceProject/android-audiobook/pull/58 (not a dependency)

No

No

Yes (@jdempcy)

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -271,10 +271,9 @@
       </c:changes>
     </c:release>
     <c:release date="2023-03-14T23:06:06+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
-      <c:changes/>
-    </c:release>
-    <c:release date="2023-03-14T23:06:06+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.1">
+      <c:changes>
         <c:change date="2023-03-30T00:00:00+00:00" summary="Save reading position of audiobooks on play, pause, and stop (which also includes when seeking and changing chapters)."/>
+      </c:changes>
     </c:release>
   </c:releases>
   <c:ticket-systems>
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -273,6 +273,9 @@
     <c:release date="2023-03-14T23:06:06+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
       <c:changes/>
     </c:release>
+    <c:release date="2023-03-14T23:06:06+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.1">
+        <c:change date="2023-03-30T00:00:00+00:00" summary="Save reading position of audiobooks on play, pause, and stop (which also includes when seeking and changing chapters)."/>
+    </c:release>
   </c:releases>
   <c:ticket-systems>
     <c:ticket-system default="true" id="org.nypl.jira" url="https://jira.nypl.org/browse/"/>
--- a/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
+++ b/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
@@ -649,18 +649,27 @@ class AudioBookPlayerActivity :
       }
 
       is PlayerEventPlaybackStarted -> {
-        this.savePlayerPosition(PlayerEventCreateBookmark(event.spineElement, event.offsetMilliseconds,
-          isLocalBookmark = false))
+        this.savePlayerPosition(PlayerEventCreateBookmark(
+          event.spineElement,
+          event.offsetMilliseconds,
+          isLocalBookmark = false)
+        )
       }
 
       is PlayerEventPlaybackPaused -> {
-        this.savePlayerPosition(PlayerEventCreateBookmark(event.spineElement, event.offsetMilliseconds,
-          isLocalBookmark = false))
+        this.savePlayerPosition(PlayerEventCreateBookmark(
+          event.spineElement,
+          event.offsetMilliseconds,
+          isLocalBookmark = false)
+        )
       }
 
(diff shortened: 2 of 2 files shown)


jdempcy:
I want to refactor this as well to de-duplicate the code as right now it's simply duplicating the same code with different signatures. I can explore either type casting or refactoring so that all the different signatures call the same method with the relevant data.

jdempcy:
I have refactored the code thusly, but I am running into an issue with the offsetMilliseconds as it is not defined in a PlayerEventWithSpineElement event:

```

  private fun savePlayerPosition(event: PlayerEventCreateBookmark) {
    savePlayerPosition(event as PlayerEvent.PlayerEventWithSpineElement)
  }

  private fun savePlayerPosition(event: PlayerEventPlaybackStarted) {
    savePlayerPosition(event as PlayerEvent.PlayerEventWithSpineElement)
  }

  private fun savePlayerPosition(event: PlayerEventPlaybackPaused) {
    savePlayerPosition(event as PlayerEvent.PlayerEventWithSpineElement)
  }

  private fun savePlayerPosition(event: PlayerEventPlaybackStopped) {
    savePlayerPosition(event as PlayerEvent.PlayerEventWithSpineElement)
  }
  private fun savePlayerPosition(event: PlayerEvent.PlayerEventWithSpineElement) {
    try {
      val bookmark = Bookmark.AudiobookBookmark.create(
        opdsId = this.parameters.opdsEntry.id,
        location = PlayerPosition(
          title = event.spineElement.position.title,
          part = event.spineElement.position.part,
          chapter = event.spineElement.position.chapter,
          offsetMilliseconds = event.offsetMilliseconds
        ),
        duration = event.spineElement.duration?.millis ?: 0L,
        kind = BookmarkKind.BookmarkLastReadLocation,
        time = DateTime.now(),
        deviceID = AudioBookDevices.deviceId(this.profilesController, this.parameters.bookID),
        uri = null
      )

      if (event is PlayerEventCreateBookmark && event.isLocalBookmark) {
        lastLocalBookmark = bookmark
        this.bookmarkService.bookmarkCreateLocal(
          accountID = this.parameters.accountID,
          bookmark = bookmark
        )
      } else {
        this.bookmarkService.bookmarkCreateRemote(
          accountID = this.parameters.accountID,
          bookmark = bookmark
        )
      }

    } catch (e: Exception) {
      this.log.error("could not save player position: ", e)
    }
  }
```

I will continue looking for a solution to this. I am hoping to not have to duplicate the code for each event type.

README-CHANGES.xml
```diff
@@ -273,6 +273,9 @@
     <c:release date="2023-03-14T23:06:06+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
       <c:changes/>
     </c:release>
+    <c:release date="2023-03-14T23:06:06+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.1">
+        <c:change date="2023-03-30T00:00:00+00:00" summary="Save reading position of audiobooks on play, pause, and stop (which also includes when seeking and changing chapters)."/>
+    </c:release>
```
nunommts: I think you've made the same mistake on this PR as you did [here](https://github.com/ThePalaceProject/android-audiobook/pull/58), because this is one version ahead of the current version

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -336,7 +351,7 @@ class AudioBookPlayerActivity :
         uri = null
       )
 
-      if (event.isLocalBookmark) {
+      if (event is PlayerEventCreateBookmark && event.isLocalBookmark) {
```
nunommts, line 354: I think this approach may be a bit tricky because we are accepting an event of the type `PlayerEvent.PlayerEventWithSpineElement` which is basically the super type of the events. Which means this method would send a remote bookmark to the server on every event that is different from the `PlayerEventCreateBookmark` and we would have to get an eye on it every time we changed the `onPlayerEvent` method. 

My suggestion is:
- we keep the `savePlayerPosition(event: PlayerEventCreateBookmark)` as the only method and with no changes;
- on the `onPlayerEvent(event: PlayerEvent)` method we create a PlayerEventCreateBookmark event for every new handled event you added (`PlayerEventPlaybackStarted`, `PlayerEventPlaybackPaused`, `PlayerEventPlaybackStopped`) with the `isLocalBookmark` as false;

What's your opinion on that?

nunommts (changes requested):

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
jdempcy: Oh! That makes sense to me. I'll make that change and follow up!

jdempcy:
Now it creates new PlayerEventCreateBookmark events and that is a much cleaner solution, so I just want to test and make sure it is working - but I like this way forward! Great advice & guidance, Nuno! I'll follow up and tag you when it's ready for review after testing.

nunommts:
@jdempcy That's really good news! Whenever you finish editing this PR please just update the `build_libraries.toml` file by running the `git submodule update` command so it can get the newest release on the Audiobook API

jdempcy:
@nunommts I ran `git submodule update` and verified that it all still works and is creating the bookmark as expected, so we should be good to go! Just let me know if there is anything else to change, otherwise let's merge it!

nunommts (approved):
Looks good!

PR #174, opened 2023-03-27
Added support to the new Audible TOC

nunommts:
**What's this do?**
This PR adds support to the new Audible TOC by adapting the bookmark saving and loading to the newest bookmark fields.

There's a [feature request](https://www.notion.so/Update-Android-audiobook-player-to-support-Audible-TOC-0307c14b2f404b708d5744e6473d3187) to implement changes to the new Audible TOC on the audiobook submodule and that requires some changes on the Palace app too.

Follow the testing steps of [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/52)

[This PR](https://github.com/ThePalaceProject/android-audiobook/pull/52) needs to be merged first. Also, [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/59) also needs to be merged first to support the latest offset changes.

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-bookmarks-api/src/main/java/org/nypl/simplified/bookmarks/api/BookmarkAnnotationsJSON.kt
+++ b/simplified-bookmarks-api/src/main/java/org/nypl/simplified/bookmarks/api/BookmarkAnnotationsJSON.kt
@@ -383,7 +383,8 @@ object BookmarkAnnotationsJSON {
     val objectNode = objectMapper.createObjectNode()
     objectNode.put("@type", "LocatorAudioBookTime")
     objectNode.put("chapter", bookmark.location.chapter)
-    objectNode.put("time", bookmark.location.offsetMilliseconds)
+    objectNode.put("startOffset", bookmark.location.startOffset)
+    objectNode.put("time", bookmark.location.currentOffset)
     objectNode.put("part", bookmark.location.part)
     objectNode.put("title", bookmark.location.title.orEmpty())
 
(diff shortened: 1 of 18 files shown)


simplified-bookmarks-api/src/main/java/org/nypl/simplified/bookmarks/api/BookmarkAnnotationsJSON.kt
```diff
@@ -383,7 +383,8 @@ object BookmarkAnnotationsJSON {
     val objectNode = objectMapper.createObjectNode()
     objectNode.put("@type", "LocatorAudioBookTime")
     objectNode.put("chapter", bookmark.location.chapter)
-    objectNode.put("time", bookmark.location.offsetMilliseconds)
+    objectNode.put("startOffset", bookmark.location.startOffset)
+    objectNode.put("time", bookmark.location.currentOffset)
```
ray-lee, line 387: When serializing, `time` is set here to be `currentOffset`. But this doesn't seem to be what is expected by the deserialization code below. Or am I missing something?

simplified-bookmarks-api/src/main/java/org/nypl/simplified/bookmarks/api/BookmarkAnnotationsJSON.kt
```diff
@@ -405,9 +406,12 @@ object BookmarkAnnotationsJSON {
     val obj =
       JSONParserUtilities.checkObject(null, node)
 
+    val startOffset = JSONParserUtilities.getIntegerDefault(obj, "startOffset", 0).toLong()
+
     return PlayerPosition(
       chapter = JSONParserUtilities.getIntegerDefault(obj, "chapter", 0),
-      offsetMilliseconds = JSONParserUtilities.getInteger(obj, "time").toLong(),
+      startOffset = startOffset,
+      currentOffset = JSONParserUtilities.getInteger(obj, "time").toLong() - startOffset,
```
ray-lee, line 414: When deserializing, it looks like `time` is expected to be the offset from the start of the track (which makes sense, to be backward compatible with existing bookmarks). So to get `currentOffset`, the `startOffset` value (if any) is subtracted, which also makes sense. But then, why isn't `startOffset` added on to `currentOffset` when serializing `time`?

ray-lee (commented):
There seems to be a mismatch between the bookmark serialization and deserialization code. Is this a bug? I might just not be understanding it.

nunommts:
@ray-lee I've created the audiobook PR (Also, [this PR](https://github.com/ThePalaceProject/android-audiobook/pull/59) also needs to be merged first to support the latest offset changes.) and updated this one.

ray-lee (approved):
Looks good!

PR #177, opened 2023-04-04
Changed position and text of preview button

nunommts:
**What's this do?**
This PR updates the position of the preview button so it can be side by side with the other button options. This PR also hides the preview button in case the user already has the full book. Finally, the PR also changes the label "Read Preview" to "Read Sample"

There's a [feature request](https://www.notion.so/Don-t-show-the-preview-button-after-the-user-gets-the-full-book-8b1caaf026c04e769361184f9f77415b) to update the preview button position, text and displaying logic to match the iOS version of the app.

Open the Palace app
Open the "Palace Marketplace Integration Library"
Open the book "Love, Penelope" by Joanne Rocklin
Confirm the "Read Preview" button now says "Read Sample"
Get the book
Confirm the "Read Sample" button is no longer there

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -270,8 +270,11 @@
         <c:change date="2023-02-28T00:00:00+00:00" summary="Moved the 'Remove' button after the 'Get' button on the Reservations screen."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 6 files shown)


ray-lee (approved):
Looks good.

PR #178, opened 2023-04-11
Fixed WebView audiobook preview not pausing when earphones are unplugged

nunommts:
**What's this do?**
This PR runs a Javascript code block when the earphones are unplugged while the audiobook preview is being played, so the WebView can get that action and pause the preview.

There's a [ticket with a bug](https://www.notion.so/lyrasis/Android-Audiobook-samples-keep-playing-after-headphones-are-removed-2f32122e07e24bea8024716a73954634) saying that for some audiobooks the preview is not stopped when the earphones are unplugged.

Open the Palace app
Select "OverDrive Integration Test Library"
Open any audiobook
Click on "Play Sample"
Confirm the layout and commands are similar to [this](https://user-images.githubusercontent.com/79104027/231117054-c97eff43-49c5-412d-a812-81ea8a8edffa.png) (meaning it's a preview on a WebView)
Plug your earphones and start playing the preview
Unplug your earphones and confirm the preview is automatically paused

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-viewer-preview/src/main/assets/preview_player_commands.js
+++ b/simplified-viewer-preview/src/main/assets/preview_player_commands.js
@@ -1,0 +1,7 @@
+function pausePlayer() {
+    const button = document.getElementsByClassName('playback-toggle halo')[0];
+    const event = document.createEvent('HTMLEvents');
+
+    event.initEvent('click', true, true);
+    button.dispatchEvent(event);
+}
--- a/simplified-viewer-preview/src/main/assets/preview_player_commands.js
+++ b/simplified-viewer-preview/src/main/assets/preview_player_commands.js
@@ -1,7 +1,11 @@
-function pausePlayer() {
-    const button = document.getElementsByClassName('playback-toggle halo')[0];
-    const event = document.createEvent('HTMLEvents');
+PalacePreviewPlayerCommands = {
+  pausePlayer: () => {
+    const button = document.querySelector('button[aria-label="Pause"]');
+    if (button) {
+      const event = document.createEvent('HTMLEvents');
 
-    event.initEvent('click', true, true);
-    button.dispatchEvent(event);
-}
+      event.initEvent('click', true, true);
+      button.dispatchEvent(event);
+    }
+  }
+};
--- a/simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
+++ b/simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
@@ -3,6 +3,7 @@ package org.nypl.simplified.viewer.preview
 import android.content.IntentFilter
 import android.media.AudioManager
 import android.os.Bundle
+import android.util.Base64
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -17,12 +18,16 @@ import androidx.core.view.WindowInsetsControllerCompat
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import org.nypl.simplified.webview.WebViewUtilities
+import java.net.URLEncoder
+import java.nio.charset.Charset
 
 class BookPreviewEmbeddedFragment : Fragment() {
 
   companion object {
     private const val BUNDLE_EXTRA_URL =
       "org.nypl.simplified.viewer.preview.BookPreviewEmbeddedFragment.url"
+
+    private const val UTF_8 = "UTF-8"
 
     fun newInstance(url: String): BookPreviewEmbeddedFragment {
       return BookPreviewEmbeddedFragment().apply {
@@ -84,7 +89,7 @@ class BookPreviewEmbeddedFragment : Fragment() {
 
   private fun configureWebView() {
     webView.apply {
-      webViewClient = WebViewClient()
+      webViewClient = CustomWebViewClient()
       webChromeClient = CustomWebChromeClient()
       settings.allowFileAccess = true
       settings.javaScriptEnabled = true
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
   }
 
   private fun pauseWebViewPlayer() {
-    webView.loadUrl(
-      "javascript:(function(){" +
-        "button=document.getElementsByClassName('playback-toggle halo')[0];" +
-        "event=document.createEvent('HTMLEvents');" +
-        "event.initEvent('click',true,true);" +
-        "button.dispatchEvent(event);" +
-        "})()"
-    )
+    webView.evaluateJavascript("pausePlayer()", null)
+  }
+
+  private inner class CustomWebViewClient : WebViewClient() {
+    override fun onPageFinished(view: WebView?, url: String?) {
+      injectJavaScript()
+      super.onPageFinished(view, url)
+    }
+
+    private fun injectJavaScript() {
+      try {
+        val inputStream = requireActivity().assets.open("preview_player_commands.js")
+        val buffer = ByteArray(inputStream.available())
+        inputStream.read(buffer)
+        inputStream.close()
+
+        val encodedContent = URLEncoder.encode(String(buffer, Charset.forName(UTF_8)), UTF_8)
+          .replace("+", "%20")
+        val encodedString = Base64.encodeToString(encodedContent.toByteArray(), Base64.NO_WRAP)
+
+        webView.evaluateJavascript(
+          "javascript:(function() {" +
+            "const parent = document.getElementsByTagName('head').item(0);" +
+            "const script = document.createElement('script');" +
+            "script.type = 'text/javascript';" +
+            "script.innerHTML = decodeURIComponent(window.atob('" + encodedString + "'));" +
+            "parent.appendChild(script)" +
+            "})()",
+          null
+        )
+      } catch (e: Exception) {
+        e.printStackTrace()
+      }
+    }
   }
 
   private inner class CustomWebChromeClient : WebChromeClient() {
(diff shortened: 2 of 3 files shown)


simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -80,7 +96,18 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
 
-  inner class CustomWebChromeClient : WebChromeClient() {
+  private fun pauseWebViewPlayer() {
+    webView.loadUrl(
+      "javascript:(function(){" +
+        "button=document.getElementsByClassName('playback-toggle halo')[0];" +
+        "event=document.createEvent('HTMLEvents');" +
+        "event.initEvent('click',true,true);" +
+        "button.dispatchEvent(event);" +
+        "})()"
```
ray-lee, line 106: I have a few concerns about this JavaScript:
- `getElementsByClassName('playback-toggle halo')` could be empty, if the player implementation ever changes. This would cause a JavaScript error, which might break the player.
- The `button` and `event` variables are global, which could break the player if it happens to read/write global variables of the same name. (Use `const` to declare them as local.)
- This only works for one version of the player for one publisher. We'll probably see different players from different publishers, and different versions of the player from each publisher. The JavaScript will eventually need to get longer and more complicated, to try to detect what kind of player it's in, and do the right thing. At that point, writing the code by building a Kotlin string is going to be hard.

I would suggest trying something like this:
1. Put the JavaScript function definition in a .js file.
2. When the player is loaded in the WebView, evaluate some JavaScript that adds a `script` tag to the page, with a `src` that loads the JavaScript from the .js file. Then the function will be defined.
3. In this `pauseWebViewPlayer` function, evaluate some JavaScript that calls the function defined in the .js file. That way, the function body can be long and complicated, and all we do here is call the function.

Does that make sense?  I can help with the JavaScript parts.

nunommts: @ray-lee Thank you for your notes, specially the ones regarding JavaScript! I did try to use the `getElementsByClassName` with a random name to see how it would behave and nothing happened. So it didn't stop the player but it also didn't cause any issues. That's why I kept it like that.

Yes, I think that made sense to me. I'll try to apply those changes and I'll reach out to you if I need any help with the JavaScript parts! Thanks!

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -80,7 +96,18 @@ class BookPreviewEmbeddedFragment : Fragment() {
     }
   }
 
-  inner class CustomWebChromeClient : WebChromeClient() {
+  private fun pauseWebViewPlayer() {
+    webView.loadUrl(
```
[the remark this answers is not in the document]
nunommts: Done!

simplified-viewer-preview/src/main/assets/preview_player_commands.js
```diff
@@ -0,0 +1,7 @@
+function pausePlayer() {
+    const button = document.getElementsByClassName('playback-toggle halo')[0];
```
ray-lee, line 2: I think it would be more robust to use this selector:
```
const button = document.querySelector('button[aria-label="Pause"]');
```
This makes sure that what you select is a button, so that sending a click event can be expected to do something. It also ensures that the intent of the button is to pause; otherwise, if the audio is not currently playing, you would toggle it, and start playing.

I would also guard against null in case the button doesn't exist. I'd rather not let it silently error, even if it doesn't break anything.

```
const button = document.querySelector('button[aria-label="Pause"]');

if (button) {
  // ...
}
```

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
+    private fun injectJavaScript() {
+      try {
+        val inputStream = requireActivity().assets.open("preview_player_commands.js")
+        val buffer = ByteArray(inputStream.available())
+        inputStream.read(buffer)
+        inputStream.close()
+
+        val encodedContent = URLEncoder.encode(String(buffer, Charset.forName(UTF_8)), UTF_8)
+          .replace("+", "%20")
+        val encodedString = Base64.encodeToString(encodedContent.toByteArray(), Base64.NO_WRAP)
```
ray-lee, line 123: Is there something you can do with [WebViewAssetLoader](https://developer.android.com/develop/ui/views/layout/webapps/load-local-content) to make this more elegant? It would be nice to do `<script src="{url to load file from assets}" />` instead of all this. This is ok if that doesn't work, but I wanted to make sure you tried it.

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
+
+        webView.evaluateJavascript(
+          "javascript:(function() {" +
+            "const parent = document.getElementsByTagName('head').item(0);" +
```
ray-lee, line 127: This can just be `document.head`. The head is special in that `document.head` will always exist, even if there is no explicit `<head>` element in the HTML.

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
+            "const script = document.createElement('script');" +
+            "script.type = 'text/javascript';" +
```
ray-lee, line 129: `type` can be omitted, because the default is `text/javascript`.

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
+            "script.innerHTML = decodeURIComponent(window.atob('" + encodedString + "'));" +
```
ray-lee, line 130: See above. If we could just set `script.src` here, that would be nicer.

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
+        val buffer = ByteArray(inputStream.available())
+        inputStream.read(buffer)
+        inputStream.close()
+
+        val encodedContent = URLEncoder.encode(String(buffer, Charset.forName(UTF_8)), UTF_8)
+          .replace("+", "%20")
+        val encodedString = Base64.encodeToString(encodedContent.toByteArray(), Base64.NO_WRAP)
+
+        webView.evaluateJavascript(
+          "javascript:(function() {" +
```
ray-lee, line 126: I don't think you need the `javascript:` here. I'm surprised that doesn't cause an error.

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
+            "const parent = document.getElementsByTagName('head').item(0);" +
+            "const script = document.createElement('script');" +
+            "script.type = 'text/javascript';" +
+            "script.innerHTML = decodeURIComponent(window.atob('" + encodedString + "'));" +
+            "parent.appendChild(script)" +
```
ray-lee, line 131: You remove the `parent` variable to simplify, and just do `document.head.appendChild(script)`.

simplified-viewer-preview/src/main/java/org/nypl/simplified/viewer/preview/BookPreviewEmbeddedFragment.kt
```diff
@@ -97,14 +102,40 @@ class BookPreviewEmbeddedFragment : Fragment() {
...
   private fun pauseWebViewPlayer() {
-    webView.loadUrl(
-      "javascript:(function(){" +
-        "button=document.getElementsByClassName('playback-toggle halo')[0];" +
-        "event=document.createEvent('HTMLEvents');" +
-        "event.initEvent('click',true,true);" +
-        "button.dispatchEvent(event);" +
-        "})()"
-    )
+    webView.evaluateJavascript("pausePlayer()", null)
```
ray-lee, line 105: Change to `PalacePreviewPlayerCommands.pausePlayer()` to be more unique.

ray-lee (commented):
@nunommts I think this is the right idea. I mostly made suggestions about JavaScript details that I wouldn't expect you to know. I also had a question about whether you had tried setting `script.src` instead of setting `innerHTML` with all the base64 stuff.

nunommts:
@ray-lee I've tried to apply all the changes you suggested. I'm just a bit uncertain if you meant these changes on the script injection to the WebView. I couldn't properly use the `WebViewAssetLoader` but I think this new way is a bit more clean. Just let me know if there's anything I can still improve/fix.

ray-lee (approved):
This looks good, thanks!

PR #182, opened 2023-05-02
Feature/add badge to holds tab

nunommts:
**What's this do?**
This PR adds logic to make automatic requests to retrieve the number of loans and to show/hide/update a badge with the available holds next to the "Holds" tab icon.

There's a [feature request](https://www.notion.so/lyrasis/Add-badge-number-to-indicate-when-a-reservation-is-available-892d0e77b27a45db8abfa8e9bb5a24b1) to add this badge and repeated automatic requests from time to time to make the app similar to the iOS version.

Open the Palace app in two different devices and with two different accounts for the LYRASIS Reads
Search for the audiobook "In Defense of Open Society" (or any other book / audiobook with just one available copy) on both devices
On one device, get the book
On the other device, click on "Reserve"
Go to the "Holds" tab and confirm there's a title in there
On the first device, return the book
On the second device, close and reopen the app and confirm a badge with the number "1" appears on the "Holds" tab
Go to the "Holds tab" on the second device, get the book and confirm the badge is now gone

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -270,7 +270,7 @@
         <c:change date="2023-02-28T00:00:00+00:00" summary="Moved the 'Remove' button after the 'Get' button on the Reservations screen."/>
       </c:changes>
     </c:release>
-    <c:release date="2023-04-27T09:01:38+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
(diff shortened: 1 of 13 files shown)


ray-lee (approved):
Looks good.

PR #183, opened 2023-05-09
Removed old pdf reader

nunommts:
**What's this do?**
This PR removes all the references to the old PDF reader as well as the option to enable/disable it from the debug settings.

There's a [feature request](https://www.notion.so/lyrasis/Android-Remove-legacy-PDF-reader-abfcb60bfa234c57851fd1d64738a0f6) saying to remove this old PDF reader as the new one has been launched into production and hasn't had any major issues.

Open the Palace app
Select "LYRASIS Reads" Library
Open any PDF
Confirm it's being correctly opened

None

Yes

No

Tested by @nunommts


nunommts:
@ray-lee I thought of renaming the "pdfjs" and its references to "pdf" as it's now the only pdf reader, but I didn't so the PR wouldn't become too complex. What do you think? Should we rename it?

ray-lee:
@nunommts I think it makes sense to rename it, but let's do it after this PR gets merged, so it's easier for me to review.

ray-lee (approved):
Looks good!

PR #185, opened 2023-05-10
Bluetooth audio controls added support

jdempcy:
**What's this do?**
Improved support of Bluetooth audio controls. Now we handle both play and pause because some headsets will alternate sending play and pause messages. Either play or pause plays/unpauses, so they do the same thing, which is because some headsets only send play or pause messages. For instance, the Bose headset tested will always send play, so the play message toggles between playing and pause. The Sony headset tested will alternate sending play and pause messages, so those work as expected as well to toggle play and pause.

Additionally, fast forward/rewind are now supported and will seek ahead or behind in 15-second increments for as long as the fast forward/rewind keys are held down (or the touchdown continues in the case of touch events).

Improved support of more devices to control play/pause with Bluetooth, as well as adding support for fast forward/rewind.
(Notion)[https://www.notion.so/lyrasis/Bluetooth-headphone-controls-don-t-work-158070702bf04351add0451a20023c2a]

Use as many different Bluetooth headsets or media controls as possible to ensure that audiobooks can play, pause, move to previous track, next track, fast forward, and rewind.

For instance, on the Sony WH-H900N headset you can perform these operations with the following gestures:

Slow double tap (0.4s apart): Play, pause
Swipe forward: next track
Swipe back: start track over, or previous track if already at the beginning of the track
Swipe forward and hold: seek forward (continues to seek in 15s increments while holding)
Swipe back and hold: seek backward (continues to seek in 15s increments while holding)

On other devices there may be hardware buttons or other gestures. But these are the features to test, using whichever devices are available.

The PR in android-audiobook (here) [https://github.com/ThePalaceProject/android-audiobook/pull/66] must be approved and merged first - this will remain a draft PR until that occurs.

Yes

No

Yes - in android-audiobook using the included audiobooks in the demo activity. Integration testing in android-core will occur once it is merged using additional audiobooks.


jdempcy:
This will remain a draft until the PR in android-audiobook [here](https://github.com/ThePalaceProject/android-audiobook/pull/66) is merged.

PR #187, opened 2023-06-16
Fixed audiobook bookmarks being incorrectly displayed and failing to be deleted

nunommts:
**What's this do?**
This PR fixes audiobook bookmarks being incorrectly displayed by hiding those who don't have a valid url. Also, this PR deletes some unnecessary code that could have been causing the bookmarks to be properly deleted.

There's [a ticket with a bug](https://ebce-lyrasis.atlassian.net/browse/PP-48) saying that some audiobook bookmarks are failing to be deleted and then they seem to be duplicated while being displayed to the user.

Open the app
Get an audiobook
Tap “Listen”
Add several bookmarks
Open TOC - Bookmarks
Delete all the bookmarks
Close the book
Tap “Listen” again
Open TOC - Bookmarks
Confirm the bookmarks are not duplicated and the ones deleted are not there anymore

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
+++ b/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
@@ -272,6 +272,45 @@ class AudioBookPlayerActivity :
     return entry.authors.first()
   }
 
+  override fun onDestroy() {
+    this.log.debug("onDestroy")
+    super.onDestroy()
+
+    /*
+     * We set a flag to indicate that the activity is currently being destroyed because
+     * there may be scheduled tasks that try to execute after the activity has stopped. This
+     * flag allows them to gracefully avoid running.
+     */
+
+    this.destroying = true
+
+    /*
+     * Cancel downloads, shut down the player, and close the book.
+     */
+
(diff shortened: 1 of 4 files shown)


simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -272,55 +272,6 @@ class AudioBookPlayerActivity :
     return entry.authors.first()
   }
 
-  override fun onDestroy() {
```
ray-lee, line 275: Why is this being removed?

nunommts: Sorry, I should've included that on the description. It's being removed because thanks to the latest changes on the player events, it creates a remote bookmark when it's paused. So, if the _onDestroy_ was implemented it would create it twice because the player is paused before being destroyed.

nunommts: You're right, my mistake! I only meant to delete that part but somehow deleted everything. Reverting it now

ray-lee (approved):
Looks good.

PR #188, opened 2023-06-22
Updated Readium version

nunommts:
**What's this do?**
This PR updates the Readium library version

So [this PR](https://github.com/ThePalaceProject/android-r2/pull/20) can be tested


ray-lee (approved):
Looks good.

PR #190, opened 2023-06-28
Updated library card creation flow

nunommts:
**What's this do?**
This PR updates the library card creation button click in two main ways:
1 - Before creating a card, the app confirms if the user granted access to their location. If not, the location is requested and, in this case, if it's rejected forever, the UI tells the user they can't create a card unless the location access is granted and how to do so. If it's granted or just rejected, the UI remains the same explaining the location is required;
2 - After granting the location or having it granted already, a WebView is launched with the library's card creation url with two new query parameters, corresponding to the user's current latitude and longitude.

There's a [feature ticket](https://ebce-lyrasis.atlassian.net/browse/PP-81) saying the card creation should now be done within a WebView and not on the phone's browser and another [feature ticket](https://ebce-lyrasis.atlassian.net/browse/PP-82) saying the latitude and longitude should be passed alongside the url. There's also a [feature ticket](https://ebce-lyrasis.atlassian.net/browse/PP-12) to add the location permission prompt.

Open the Palace app
Select a library that allows Virtual Library Card creation (p.e., Bethel Public Library)
Confirm there's [information about retrieving the user's current location](https://github.com/ThePalaceProject/android-core/assets/79104027/c90b1be0-4c2e-44ec-92ae-5801368b0af3)
Click on "Create Card"
Confirm the app requests the user's location
Click on "Reject"
Confirm the UI remains the same
Click on "Create Card" again and reject it again but this time select the "do not ask again" option
Confirm the [UI changed](https://github.com/ThePalaceProject/android-core/assets/79104027/1ba01baa-87b8-458e-ba99-f943ff1b96db)
Click on "Open Settings"
Go to permissions and grant the location
Return to the Palace app
Confirm the UI changed again
Click on "Create Card"
Confirm a [WebView is opened with the card creation url](https://github.com/ThePalaceProject/android-core/assets/79104027/5de27b19-8ecf-4ae2-8798-bc16df147a63)
Go back

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountDetailFragment.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountDetailFragment.kt
@@ -1,8 +1,11 @@
 package org.nypl.simplified.ui.accounts
 
+import android.Manifest
 import android.content.Intent
+import android.content.pm.PackageManager
 import android.net.Uri
 import android.os.Bundle
+import android.provider.Settings
 import android.view.View
 import android.view.ViewGroup
 import android.webkit.URLUtil
@@ -11,12 +14,16 @@ import android.widget.ImageView
 import android.widget.ProgressBar
 import android.widget.TextView
 import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.widget.SwitchCompat
 import androidx.constraintlayout.widget.ConstraintLayout
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
 import androidx.core.os.bundleOf
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
+import com.google.android.gms.location.LocationServices
 import com.io7m.junreachable.UnimplementedCodeException
 import com.io7m.junreachable.UnreachableCodeException
 import io.reactivex.disposables.CompositeDisposable
@@ -88,6 +95,10 @@ class AccountDetailFragment : Fragment(R.layout.account) {
     }
   )
 
+  private val fusedLocationClient by lazy {
+    LocationServices.getFusedLocationProviderClient(requireActivity())
+  }
+
   private val cardCreatorLauncher: ActivityResultLauncher<CardCreatorContract.Input>? =
     services.optionalService(CardCreatorServiceType::class.java)
       ?.getCardCreatorContract()
@@ -95,6 +106,20 @@ class AccountDetailFragment : Fragment(R.layout.account) {
 
   private val imageLoader: ImageLoaderType =
     services.requireService(ImageLoaderType::class.java)
+
+  private val locationPermissions = arrayOf(
+    Manifest.permission.ACCESS_FINE_LOCATION,
+    Manifest.permission.ACCESS_COARSE_LOCATION,
+  )
+
+  private val locationPermissionCallback =
+    registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) { results ->
+      if (results.values.all { it }) {
+        openCardCreatorWebView()
+      } else {
+        setLayoutAccordingToLocationPermissions()
+      }
+    }
 
   private lateinit var accountCustomOPDS: ViewGroup
   private lateinit var accountCustomOPDSField: TextView
@@ -122,6 +147,7 @@ class AccountDetailFragment : Fragment(R.layout.account) {
   private lateinit var settingsCardCreator: ConstraintLayout
   private lateinit var signUpButton: Button
   private lateinit var signUpLabel: TextView
+  private lateinit var signUpMessage: TextView
   private lateinit var toolbar: NeutralToolbar
 
   private val imageButtonLoadingTag = "IMAGE_BUTTON_LOADING"
@@ -191,6 +217,8 @@ class AccountDetailFragment : Fragment(R.layout.account) {
       view.findViewById(R.id.accountCardCreatorSignUp)
     this.signUpLabel =
       view.findViewById(R.id.accountCardCreatorLabel)
+    this.signUpMessage =
+      view.findViewById(R.id.accountCardCreatorMessage)
     this.settingsCardCreator =
       view.findViewById(R.id.accountCardCreator)
 
@@ -236,6 +264,14 @@ class AccountDetailFragment : Fragment(R.layout.account) {
     this.viewModel.accountSyncingSwitchStatus.observe(this.viewLifecycleOwner) { status ->
       this.reconfigureBookmarkSyncingSwitch(status)
     }
+  }
+
+  override fun onResume() {
+    super.onResume()
+
+    // update the layout here because the user may have granted or denied the location permission
+    // outside the app
+    setLayoutAccordingToLocationPermissions()
   }
 
   private fun reconfigureBookmarkSyncingSwitch(status: BookmarkSyncEnableStatus) {
@@ -332,28 +368,6 @@ class AccountDetailFragment : Fragment(R.layout.account) {
     }
   }
 
-  private fun openCardCreator() {
-    val cardCreatorURI = this.viewModel.account.provider.cardCreatorURI
-    if (cardCreatorURI != null) {
-      if (cardCreatorURI.scheme == this.nyplCardCreatorScheme) {
-        if (cardCreatorLauncher != null) {
-          cardCreatorLauncher.launch(
-            CardCreatorContract.Input(
-              this.authenticationViews.getBasicUser().value.trim(),
-              this.viewModel.account.loginState is AccountLoggedIn
-            )
-          )
-        } else {
-          // We rely on [shouldSignUpBeEnabled] to have disabled the button
-          throw UnreachableCodeException()
-        }
-      } else {
-        val webCardCreator = Intent(Intent.ACTION_VIEW, Uri.parse(cardCreatorURI.toString()))
-        this.startActivity(webCardCreator)
-      }
-    }
-  }
-
   override fun onStart() {
     super.onStart()
 
@@ -373,7 +387,17 @@ class AccountDetailFragment : Fragment(R.layout.account) {
      * Launch Card Creator
      */
 
-    this.signUpButton.setOnClickListener { this.openCardCreator() }
+    this.signUpButton.setOnClickListener {
+      if (!isLocationPermissionGranted()) {
+        if (shouldShowRationale()) {
+          requestLocationPermissions()
+        } else {
+          openAppSettings()
+        }
+      } else {
+        openCardCreatorWebView()
+      }
+    }
 
     /*
      * Configure the bookmark syncing switch to enable/disable syncing permissions.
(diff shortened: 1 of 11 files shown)


ray-lee (commented):
This isn't quite doing what I expect. If I have Location permission set to Ask every time, I get the prompt that says "You have disabled location services for this app," and the Open Settings button. Shouldn't I get the other prompt, with the Create Card button?

In order to see the Create Card button, I need to change the permission to Allow only while using the app.

https://github.com/ThePalaceProject/android-core/assets/1395885/40e41540-db1b-490f-bcf1-3f0e58e9b4ed

This is on Android 13.

nunommts:
@ray-lee Thanks! Yes, this is a bug. I'm thinking it has to be with the fact that you hadn't granted the location permission and, somehow, the "shouldShowRationale" method is returning false. I'll investigate this

ray-lee (commented):
It's still not quite doing what I expect. Here's what I tried:

- In Settings, change Location permission for Palace to Don't allow
- In Palace, open the Bethel Public Library account screen. As expected, the "please select Allow Once" message appears, with the Create Card button.
- Tap Create Card, and select "Only this time"
- Go back to the Bethel Public Library account screen. As expected, the "please select Allow Once" message appears, with the Create Card button.
- Close Palace
- Wait about a minute
- Open Palace, and go to the Bethel Public Library account screen. Unexpected: The "You have disabled location services" message now appears, with the Open Settings button. In Settings, the Location permission is still set to Ask every time.

nunommts:
@ray-lee hmm, I see. I'm facing the same issue as you are. The thing is that I'm not quite sure how we could handle this. When the app is launched for the first time, the `shouldShowRationale` returns false and that's why I've added that flag on SharedPreferences to know if the permission has been asked for the first time. However, as soon as the dialog is shown, the flag is set as `true` and it won't be `false` again. 

In this case, the issue is slightly different because when you set as "Ask Every Time" the permission hasn't been granted but the "shouldShowRationale()" somehow returns `false` as well so the layout displays the "Open Settings" message and action. If you set as "Don't Allow" and then return to the app, the "Create Card" action and message will be displayed because the "shouldShowRationale()" returns true. And if I remove the "shouldShowRationale()" related logic from the code, the user will always see the "Create Card" action when the permission is not granted, but if they somehow have pressed the "Don't Ask Again" option, nothing will be displayed and nothing will happen after pressing the "Create Card" button.

I'm not sure if this is a bug from the native Android code or not but I'm not sure how can we achieve the perfect solution for this. Do you have any ideas?

ray-lee:
I think the root of the problem is that the design we're trying to implement doesn't really follow the conventional Android workflow. Referencing https://developer.android.com/training/permissions/requesting, it seems like what we should do is something like this:

- On the login screen, above the Create Card button, show a message like "The Palace App requires permission to access your precise location in order to verify your library service area." At this point, we can't say anything about whether or not the user will be prompted for permissions when they tap the button, because we don't know if they (for example) have granted permissions permanently, or granted one-time permissions that are still active, or denied permissions permanently, all of which would result in no prompt being shown.

- When Create Card is tapped, check if the permission has been granted.
  - If so, continue.
  - If not:
    - Call shouldShowRequestPermissionRationale().
       - If true, show a dialog that explains why location permissions are needed.
       - If false, continue.
    - Request permissions.
       - If granted, get the location and open the WebView.
       - If denied: This is a theory based on my reading of StackOverflow (https://stackoverflow.com/questions/32347532/android-m-permissions-confused-on-the-usage-of-shouldshowrequestpermissionrati), but I think we could call shouldShowRequestPermissionRationale() again here. If the permission has been denied, and shouldShowRequestPermissionRationale() returns false, it implies that the user has indicated that they don't want to be asked about the permission again. But they did tap on the Create Card button, so they probably do want to enable the permission. So we could show a dialog here that prompts them to open Settings, with a button to do so.
      
How does that sound? We could run this by David and Alissa.

nunommts:
@ray-lee It's true, indeed. The UI may be a bit hard to combine with the usual flow of the permissions, like you said. I liked your idea, but I have three questions/comments:

1 - In the "Call shouldShowRequestPermissionRationale()" section, I think that if it's true we don't need to display that dialog to tell the user why we need the location permission because it's already explained on the screen. If we move forward with that there's no harm in showing "duplicate" information but we could also adapt the message above the "Create Card" button;
2 - Also in the "Call shouldShowRequestPermissionRationale()" you say that if "it's false, continue" but what does this mean? Are you suggesting to still open the WebView and let it handle the current permission state?
3 - The only issue of the last point you raised is that the "shouldShowRequestPermissionRationale()" returns false at the first time, so I think we'd still need the SharedPreferences flag on this case.

Despite this, since we'll probably need to do some changes I think we need to talk to both of them when we reach a feasible solution.

nunommts:
@ray-lee I thought about this for a while and considering everything we discussed, my proposed solution is:

* Display the "Create Card" button as it was before, without any text (or, maximum, a really short description)
* After pressing the button we check if the location permissions were granted:
- * If so, open the WebView;
- * If not, we check if we should show the rationale or if the permissions weren't already asked:
- - * If one of these is true, we display a dialog saying why we need the permissions with the "Ok" and "Cancel" options. Pressing "Ok", we ask for permissions. Pressing "Cancel" the dialog is closed without any further actions;
- - - * If we ask for permissions and they are granted, we open the WebView. If the permissions are denied, we can either close the dialog or show the dialog described below.
- - * If none of these is true, we display a dialog with a message saying why we need the permission and informing the user that they need to grant the permissions on the app settings with a "Open Settings" and a "Cancel" options. Pressing "Open Settings" redirects the user to the app settings. Pressing "Cancel", the dialog is closed without any further actions.


What do you think?

ray-lee:
> 1 - In the "Call shouldShowRequestPermissionRationale()" section, I think that if it's true we don't need to display that dialog to tell the user why we need the location permission because it's already explained on the screen. If we move forward with that there's no harm in showing "duplicate" information but we could also adapt the message above the "Create Card" button;

The reason I had both the message above the button and the dialog with the rationale is that the suggested workflow (https://developer.android.com/training/permissions/requesting) has both "2. Design your app UX to be self-explanatory" and "5b. Explain to the user why your app needs this permission". I think we're supposed to just say we need the permission first, and later provide more of an explanation, if the system thinks that the user needs to be convinced.

> 2 - Also in the "Call shouldShowRequestPermissionRationale()" you say that if "it's false, continue" but what does this mean? Are you suggesting to still open the WebView and let it handle the current permission state?

Continue to requesting the permission. This will either show the dialog to ask for permission, or if permission has already been granted, do nothing (and the next step would be to show the WebView).

 3 - The only issue of the last point you raised is that the "shouldShowRequestPermissionRationale()" returns false at the first time, so I think we'd still need the SharedPreferences flag on this case.

Right, but this call would happen only after the permission has been requested and denied, so it is guaranteed not to be the first time.

ray-lee:
> * Display the "Create Card" button as it was before, without any text (or, maximum, a really short description)

I would vote for a really short description, because I think we're supposed to have something to satisfy "2. Design your app's UX so that specific actions in your app are associated with specific runtime permissions. **Let users know which actions might require them to grant permission for your app to access private user data.**" The way I interpret this is that we just need to let users know that precise location access will be required, before they tap the button.

> * After pressing the button we check if the location permissions were granted:
>   
>   * If so, open the WebView;
>   * If not, we check if we should show the rationale or if the permissions weren't already asked:

I don't think we should care if permissions weren't already asked. We should just call shouldShowRequestPermissionRationale(), and let Android decide if we should explain why we need the permissions. I think it returns false the first time on purpose, because they expect that most users will just allow it, and they don't want to make an extra step for them. At this point we would just be following the suggested workflow (https://developer.android.com/training/permissions/requesting).
 
> we display a dialog saying why we need the permissions with the "Ok" and "Cancel" options. Pressing "Ok", we ask for permissions. Pressing "Cancel" the dialog is closed without any further actions:

This sounds good.
     
>     * If none of these is true, we display a dialog with a message saying why we need the permission and informing the user that they need to grant the permissions on the app settings with a "Open Settings" and a "Cancel" options. Pressing "Open Settings" redirects the user to the app settings. Pressing "Cancel", the dialog is closed without any further actions.

The trick for showing this dialog is to only do it if the user _will not_ be prompted again to allow the permissions, the next time they click the Create Card button. In modern versions of Android, this would only be after the second time they've denied the request. That's why I had the condition that in order to show this dialog, the permission request must have just been denied, and shouldShowRequestPermissionRationale() should return false. I think that will be the right thing, but we'll have to do some experimentation.

nunommts:
@ray-lee I'm writing a comment on one of the tickets explaining what we discussed here. There's still one thing that I'm not quite sure of what we should do:

- On the first time we go to that screen, the user hasn't granted the permission and the `shouldShowRationale()` will return false. According to the suggested flow by Android, in this case we should request for the permission. Since it's the first time, it will show the location permission dialog. However, there might be the case the user denied the permission forever, so it would also return false on permission granted and on the rationale method. In this case, we  will try to request the permission but nothing will happen. If they start pressing the "Create Card" button repeatedly, nothing will happen and it may be confusing for the user. Should we leave it like this?

nunommts:
@ray-lee I've updated both the PR and the PR's description according to Alissa's comment on [this card](https://ebce-lyrasis.atlassian.net/browse/PP-212)

ray-lee:
@nunommts Could you get #195 into this branch? I'm running into that error now.

I was able to run this once, and got this error:
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.thepalaceproject.palace, PID: 18571
    java.lang.NullPointerException: Attempt to invoke virtual method 'double android.location.Location.getLatitude()' on a null object reference
        at org.nypl.simplified.ui.accounts.AccountDetailFragment.openCardCreatorWebView$lambda-20(AccountDetailFragment.kt:1135)
        at org.nypl.simplified.ui.accounts.AccountDetailFragment.$r8$lambda$8vcKb9zrqHH1pHz_LsZk2PWn8PI(Unknown Source:0)
        at org.nypl.simplified.ui.accounts.AccountDetailFragment$$ExternalSyntheticLambda12.onSuccess(Unknown Source:6)
        at com.google.android.gms.tasks.zzm.run(com.google.android.gms:play-services-tasks@@18.0.1:1)
        at android.os.Handler.handleCallback(Handler.java:942)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:201)
        at android.os.Looper.loop(Looper.java:288)
        at android.app.ActivityThread.main(ActivityThread.java:7898)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
```

nunommts:
@ray-lee I've merged that PR into this branch but I also added some code to prevent the crash you had from happening again

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountDetailFragment.kt
```diff
@@ -96,6 +108,23 @@ class AccountDetailFragment : Fragment(R.layout.account) {
...
+  private val locationPermissionCallback =
+    registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) { results ->
+      if (results.values.all { it }) {
+        openCardCreatorWebView()
+      } else {
+        showSettingsDialog()
+      }
+
+      // the permissions were shown, so we can store that value
+      Cache(requireContext()).setLocationPermissionsAsked()
```
ray-lee, line 125: This isn't being used any more, right?

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountDetailFragment.kt
```diff
@@ -96,6 +108,23 @@ class AccountDetailFragment : Fragment(R.layout.account) {
...
+    Manifest.permission.ACCESS_FINE_LOCATION,
+    Manifest.permission.ACCESS_COARSE_LOCATION,
+  )
+
+  private val locationPermissionCallback =
+    registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) { results ->
+      if (results.values.all { it }) {
+        openCardCreatorWebView()
+      } else {
+        showSettingsDialog()
```
ray-lee, line 121: I think this is where the flow digram said that the settings dialog would only be shown if a call to shouldShowRationale() returns false, right? Otherwise this dialog will be shown before it needs to be. I would only expect it to be shown after the second time the user has denied permissions.

nunommts: Yes, you're right!

ray-lee (approved):
Looks good!

PR #191, opened 2023-07-04
Fixed bug of bookmarks not being deleted

nunommts:
**What's this do?**
When adding a bookmark, an object of the type  Bookmark was created but it was given 'null' as the URI field value which was causing the bookmark to not be able to be deleted since it needs a non-null URI in order to perform the operation on the server. However, the UI was being updated as if it was actually deleted, so when the user left the audiobook/ebook and returned, they'd still be seeing the "deleted" bookmarks. This PR fixes this bug by receiving the created bookmark from the server and saving it with a valid URI so it can be properly deleted right after being added or later. This PR also updates the UI in order to give some feedback to the user when the bookmark is being created/deleted.

There's a [report of a bug](https://ebce-lyrasis.atlassian.net/browse/PP-48) saying the bookmarks are deleted but reappear after reopening a book/audiobook.

Open the Palace app
Select "LYRASIS Reads"
Open any eBook/audiobook
Create a bookmark and confirm it's added
Go to the bookmarks and confirm it's there
Delete it
Go back and leave the book/audiobook
Reopen the book/audiobook and go to Bookmarks
Confirm there are no bookmarks
Go back and add a new bookmark
Leave the book/audiobook, reopen it and go to Bookmarks
Confirm the bookmark is there
Delete it
Go back and leave the book/audiobook
Reopen the book/audiobook and go to Bookmarks
Confirm there are no bookmarks

This [audiobook PR](https://github.com/ThePalaceProject/android-audiobook/pull/69) and this [reader PR](https://github.com/ThePalaceProject/android-r2/pull/21) need to be merged first.

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -270,7 +270,7 @@
         <c:change date="2023-02-28T00:00:00+00:00" summary="Moved the 'Remove' button after the 'Get' button on the Reservations screen."/>
       </c:changes>
     </c:release>
-    <c:release date="2023-06-26T10:23:13+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
+    <c:release date="2023-07-04T08:49:55+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
       <c:changes>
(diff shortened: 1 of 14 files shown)


ray-lee (approved):
Looks good to me!

PR #192, opened 2023-07-11
Added default LCP passphrase to be returned when the feed entry doesn't have one and the manual input is enabled

nunommts:
**What's this do?**
This PR adds an empty passphrase as a default value to return when two conditions are met: no LCP passphrase on the feed entry and the LCP passphrase manual input setting is enabled.

There's a comment on [this ticket](https://ebce-lyrasis.atlassian.net/browse/PP-13) saying that there are some LCP protected books that are failing to be downloaded on the Android app while the same books are downloaded on the iOS version

Open the Palace app
Enable the hidden libraries option on the Debug options
Go to Libraries and select _Cerberus Test Library 1_
Open any book from _EDRLab LCP Test_ lane
Download it and confirm the download is successful even if the book fails to be opened

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -270,7 +270,7 @@
         <c:change date="2023-02-28T00:00:00+00:00" summary="Moved the 'Remove' button after the 'Get' button on the Reservations screen."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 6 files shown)


ray-lee (approved):
Looks good.

PR #193, opened 2023-07-21
PP-187: Fix some bookmark issues

io7m:
Affects: https://ebce-lyrasis.atlassian.net/browse/PP-185

**What's this do?**

This corrects a number of bookmarking issues.

* The existing code leaked a thread on every controller event by creating an executor each time but not shutting it down.
* The SR2 code has been updated to publish the correct event on each page turn.
* The code now unconditionally creates a local bookmark even if saving the remote bookmark failed.
* Toast messages are not shown for page turns because this tends to result in an endless queue of toast messages.

https://ebce-lyrasis.atlassian.net/browse/PP-187

Open an ePub and verify that turning pages and creating bookmarks works. Verify that they still work even if annotation syncing is disabled.

https://github.com/ThePalaceProject/android-r2/pull/22

Nope!

Nope!

@io7m tested locally.

### DIFF
--- a/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
+++ b/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
@@ -33,6 +33,15 @@ object FluentFutureExtensions {
   }
 
   /**
+   * Apply a function `f` to the value of the current future. This is the same as #map except
+   * that the passed in value might be null.
+   */
+
+  fun <A, B> FluentFuture<A?>.mapNulllable(f: (A?) -> B): FluentFuture<B> {
+    return this.transform(Function<A?, B> { x -> f.invoke(x) }, MoreExecutors.directExecutor())
+  }
+
+  /**
    * Apply a function `f` to the value of the current future, yielding a new future.
    */
 
--- a/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
+++ b/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
@@ -37,7 +37,7 @@ object FluentFutureExtensions {
    * that the passed in value might be null.
    */
 
-  fun <A, B> FluentFuture<A?>.mapNulllable(f: (A?) -> B): FluentFuture<B> {
+  fun <A, B> FluentFuture<A?>.mapNullable(f: (A?) -> B): FluentFuture<B> {
     return this.transform(Function<A?, B> { x -> f.invoke(x) }, MoreExecutors.directExecutor())
   }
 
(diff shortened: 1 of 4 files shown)


simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Activity.kt
```diff
@@ -505,6 +499,24 @@ class Reader2Activity : AppCompatActivity(R.layout.reader2) {
     }
   }
 
+  private fun onBookmarkWasCreated(
+    possiblyRemoteBookmark: Bookmark?,
+    localBookmark: Bookmark.ReaderBookmark,
+    event: SR2Event.SR2BookmarkEvent.SR2BookmarkCreate
+  ) {
+    val savedBookmark = possiblyRemoteBookmark ?: localBookmark
```
nunommts, line 507: Instead of passing the two bookmarks as an argument and to this assignment inside the method, you could just pass one argument with this verification here, i.e., `onBoomarkWasCreated(possiblyRemoteBookmark ?: localBookmark, event)`

simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Activity.kt
```diff
@@ -422,40 +430,29 @@ class Reader2Activity : AppCompatActivity(R.layout.reader2) {
   private fun onControllerEvent(
     event: SR2Event
   ) {
-
-    val backgroundThread =
-      MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(1))
-
```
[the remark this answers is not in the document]
nunommts: Oh, nice catch!

simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
```diff
@@ -32,6 +32,15 @@ object FluentFutureExtensions {
     return this.transform(Function<A, B> { x -> f.invoke(x!!) }, MoreExecutors.directExecutor())
   }
 
+  /**
+   * Apply a function `f` to the value of the current future. This is the same as #map except
+   * that the passed in value might be null.
+   */
+
+  fun <A, B> FluentFuture<A?>.mapNulllable(f: (A?) -> B): FluentFuture<B> {
```
nunommts, line 40: Typo on `nulllable`?

nunommts (changes requested):

simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
io7m: Oops.

io7m:
Fixed those!

nunommts (approved):

PR #194, opened 2023-07-26
PP-181: Refactor deep link method

io7m:
**What's this do?**
This refactors the deep link intercept method to avoid several potential crashes.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-181

Try using a deep link after creating a library card.

None.

Nope.

Nope.

Unfortunately, I've not been able to get a link to actually trigger the app. This code change does make the original crash impossible, but somebody with a working device/link will need to try it.

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
@@ -14,6 +14,8 @@ import org.librarysimplified.services.api.Services
 import org.nypl.simplified.accounts.api.AccountID
 import org.nypl.simplified.accounts.registry.api.AccountProviderRegistryType
 import org.nypl.simplified.buildconfig.api.BuildConfigurationServiceType
+import org.nypl.simplified.deeplinks.controller.api.DeepLinksControllerType
+import org.nypl.simplified.deeplinks.controller.api.ScreenID
 import org.nypl.simplified.listeners.api.ListenerRepository
 import org.nypl.simplified.listeners.api.listenerRepositories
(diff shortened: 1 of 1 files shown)


simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
```diff
@@ -79,49 +81,77 @@ class MainActivity : AppCompatActivity(R.layout.main_host) {
...
+      val screenId =
+        when (screenRaw) {
+          null -> kotlin.run {
+            this.logger.warn("Deep link did not have a screen parameter.")
+            ScreenID.UNSPECIFIED
+          }
+          "login" -> ScreenID.LOGIN
+          else -> kotlin.run {
+            this.logger.warn("Deep link had an unrecognized screen parameter {}.", screenRaw)
+            ScreenID.UNSPECIFIED
```
nunommts, line 145: Are these `kotlin.run` needed here? Also, I think it would be better if this "login" value was declared as a `const val` inside the `companion object` block

nunommts (changes requested):

simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
io7m: Amusingly, the IDE actually inserted those. I thought it looked a little odd... They're just supposed to be block statements. I think there's an unfortunate syntactic aspect of Kotlin where you intend to declare a block but you actually declare a lambda and then up not executing anything (because the syntax is nearly identical).

I'll remove the `run` calls.

I agree about the constants too. I just copied the original code pretty directly.

nunommts (approved):

PR #195, opened 2023-07-28
PP-181: Links can be null

io7m:
**What's this do?**
Avoids crashing on a null link.

https://ebce-lyrasis.atlassian.net/browse/PP-181


simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
```diff
@@ -91,6 +91,10 @@ class MainActivity : AppCompatActivity(R.layout.main_host) {
     }
 
     pendingLink.addOnSuccessListener { linkData ->
+      if (linkData == null) {
```
nunommts, line 94: I'd suggest to do `val deepLink = linkData?.link` because that way we wouldn't more any more if conditions

nunommts (changes requested):

nunommts (approved):

PR #196, opened 2023-07-31
Transifex proof-of-concept.

ray-lee:
~This is a proof of concept. Do not merge.~

This demonstrates Transifex integration. To test:

- Change the primary language on your device to Spanish (Settings/System/Languages & input/Languages/Add a language, move Spanish to the top of the list).
- Build and run the app from this branch.
  (Note: It's normal in this build to see strings enclosed in square brackets. This indicates that the string has not been translated.)
- Open the catalog screen on a library. The header should read "Catalogar" (which is an incorrect translation, but that's beside the point). The More links on lanes should read "[[More...]]", enclosed in double brackets, since the string is untranslated.
- Open the My Books screen. The title should read "[[My Books]]", since it is untranslated in the `es/txstrings.json` file bundled with the app.
- Restart the app.
- Open the My Books screen. The title should now read "Mis libros", since an updated translation file was downloaded the previous time the app started, and this file does have a translation for "My Books".

Work remaining:
- ~Automate upload of strings to Transifex, so new strings are available to translators.~
- ~Automate download of latest translations for bundling with the app.~
- Enable Transifex in all activities (currently only `MainActivity` and `Reader2Activity` are enabled).
- Locate strings that aren't being translated and deal with them.
- Clean up.


io7m:
I think this is ready to go. I did the following:

* Added Transifex support for the remaining activities
* Moved the Transifex initialization into its own class
* Put Transifex into the dependency catalog instead of the declarations being inlined in build scripts
* Tried it and failed to find any strings that weren't translated

I didn't:

* Automate uploads/downloads. The problem is that this is going to involve build system work, and we really need to redo the build systems in all of the Android projects first; they're creaking under their own weight of technical debt.

nunommts (approved):
This looks good!

PR #197, opened 2023-08-02
Fixed crash after setting an audiobook timer more than once in a short time

nunommts:
**What's this do?**
This PR adds a try-catch to the holds call on the MainFragmentViewModel because the app was crashing everytime the call was interrupted.

There's a [ticket reporting a bug](https://ebce-lyrasis.atlassian.net/browse/PP-116) that when an audiobook timer was set more than once in a short time the app was crashing. This was one of the reasons causing the call interruption and consequent crash, but there were more since this crash was reported several times on Crashlytics.

Open the Palace app
Select any library
Open any audiobook
Start playing it
Change the playback speed
Set a timer
Set a different time right after setting the first one
Confirm the app is not crashing

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
@@ -52,7 +52,7 @@ class MainFragmentViewModel(
     UnicastWorkSubject.create()
   val registryEvents: UnicastWorkSubject<BookStatusEvent> =
     UnicastWorkSubject.create()
-  val deepLinkEvents: UnicastWorkSubject<org.nypl.simplified.deeplinks.controller.api.DeepLinkEvent> =
+  val deepLinkEvents: UnicastWorkSubject<DeepLinkEvent> =
     UnicastWorkSubject.create()
   val bookHoldEvents: UnicastWorkSubject<BookHoldsUpdateEvent> =
     UnicastWorkSubject.create()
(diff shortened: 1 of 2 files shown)


simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
```diff
@@ -150,19 +150,24 @@ class MainFragmentViewModel(
...
-          .size
-
-        bookRegistry.updateHolds(
-          numberOfHolds = numberOfHolds
-        )
+
+        try {
+          val numberOfHolds = getHoldsFuture
+            ?.get()
+            ?.entriesInOrder
```
io7m, line 157: On what thread does this call happen?

io7m: The reason I ask is that it's rarely a good idea to call `get()` on a future directly as it's a blocking call. If this happens on the UI thread... That's bad.

nunommts: This is running on a background thread, but the `initializeHoldsCallTimer()` was being called while that call was running and since it interrupts it, the app was crashing because of a `InterruptedException`

nunommts: That's true, but the `getHoldsFuture` variable is declared as nullable and that's why the `?` are being used.

io7m: Ah, I see. I didn't realize it was declared above as a field in the class.

I don't like how this viewmodel is structured, but this PR doesn't make it any worse!

io7m (approved):

simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
nunommts: Refactoring time is always a good time :D

PR #198, opened 2023-08-02
Fixed crash when saving a bookmark on the server

nunommts:
**What's this do?**
This PR adds a try-catch to two calls to the server: deleting and saving a bookmark. This is done to prevent the call from reaching its defined timeout and crashing the app.

There are two tickets ([here](https://ebce-lyrasis.atlassian.net/browse/PP-218) and [here](https://ebce-lyrasis.atlassian.net/browse/PP-219)) reporting a crash that was happening after saving a bookmark to the server. Even though this [ticket's crash reproducing steps](https://ebce-lyrasis.atlassian.net/browse/PP-219) aren't related with the bookmarks the reason why it's crashing is because there's a timeout while saving the bookmark on the server (because it automatically saves the reading position) as it can be verified on [Crashlytics](https://console.firebase.google.com/project/the-palace-project/crashlytics/app/android:org.thepalaceproject.palace/issues/a3f32a5ad570f0c9076856c2ae0bad3e?time=1689984000000:1690329599000&types=crash&versions=1.5.0-SNAPSHOT-debug%20(73974790)&sessionEventKey=64BEF5AB02B1000102F32E16824C5518_1837955601532794198).

Just follow both tickets' reproducing steps ([here](https://ebce-lyrasis.atlassian.net/browse/PP-218) and [here](https://ebce-lyrasis.atlassian.net/browse/PP-219))

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -270,7 +270,7 @@
         <c:change date="2023-02-28T00:00:00+00:00" summary="Moved the 'Remove' button after the 'Get' button on the Reservations screen."/>
       </c:changes>
     </c:release>
-    <c:release date="2023-08-02T13:28:21+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
+    <c:release date="2023-08-02T17:23:07+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.5.0">
       <c:changes>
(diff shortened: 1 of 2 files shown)


io7m (approved):
Interesting!

I'm curious how long this one has been occurring... That code is very old.

PR #199, opened 2023-08-08
Feature/audiobook time tracking

nunommts:
**What's this do?**
This PR adds the feature of books time tracking (for now just for the audiobooks). The idea is to increment a value for the seconds the user is listening to an audiobook and for every 60 seconds, send the value to the server if its greater than 0. In case the server, for some reason, can't process the entries that were sent, it will warn which entries were unprocessed and they should be stored to be sent on a future request. Other than this, the entries should be sent whenever the app starts or has internet connection after being offline and this feature should be enabled in the DEV settings (it's disabled by default).

There's a [feature request](https://ebce-lyrasis.atlassian.net/browse/PP-151) for the time tracking.

Go to DEV settings and enable the "Enable Time Tracking" option
Go to the "Minotaur Test Library" and open any audiobook from the _Biblioboard_ lane

None

Yes

No

Tested by @nunommts


simplified-books-time-tracking/README.md
```diff
@@ -0,0 +1,5 @@
+org.librarysimplified.books.time.tracking
+===
+
+The `org.librarysimplified.book.time.tracking` module provides
```
io7m, line 4: Provides what?! Provides what?!?!

simplified-books-time-tracking/proguard-rules.pro
io7m, line 1: Are we using Proguard now? Seems redundant.

simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingJSON.kt
```diff
@@ -0,0 +1,146 @@
...
+      TimeTrackingInfo(
+        accountId = node.get(NODE_ACCOUNT_ID).asText(),
+        bookId = node.get(NODE_BOOK_ID).asText(),
+        libraryId = node.get(NODE_LIBRARY_ID).asText(),
+        timeEntries = timeEntries,
+        timeTrackingUri = URI(node.get(NODE_URI).asText())
+      )
+    } catch (e: Exception) {
+      logger.error("Error converting bytes from file: ", e)
+      e.printStackTrace()
```
io7m, line 142: Think that one was left in accidentally.

simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingService.kt
```diff
@@ -0,0 +1,399 @@
...
+    logger.debug(
+      "Start tracking time for account {} and book {}",
+      accountID.uuid.toString(),
+      bookId
+    )
+    val libraryFile = File(timeTrackingDirectory, accountID.uuid.toString())
+
+    // create a directory for the library
+    if (!libraryFile.exists()) {
+      libraryFile.mkdirs()
```
io7m, line 89: No need to wrap this in a conditional. `mkdirs()` is idempotent.

simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingService.kt
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    val bookFile = File(libraryFile, bookId)
+
+    // create a directory for the book inside the library
+    if (!bookFile.exists()) {
+      bookFile.mkdirs()
```
io7m, line 96: Same!

simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingService.kt
```diff
@@ -0,0 +1,399 @@
...
+      }
+    }
+  }
+
+  override fun stopTracking() {
+    logger.debug("Stop tracking playing time")
+
+    audiobookPlayingDisposable?.dispose()
+    remoteStorageDisposable?.dispose()
+
```
io7m, line 172: Might consider using the `CompositeDisposable` here. It's a `Disposable` in rxjava that you can add other disposables into. When you dispose of the composite, it disposes of all of the others, even if one of the disposables raises an exception.

simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingService.kt
```diff
@@ -0,0 +1,399 @@
...
+          },
+          {
+            logger.error("Error on audiobook playing timer")
+            it.printStackTrace()
+          }
+        )
+
+    // this timer will be responsible for updating the flag to save the entries on the server
+    remoteStorageDisposable =
+      Observable.interval(60000L, TimeUnit.MILLISECONDS)
```
io7m, line 383: `(1L, TimeUnit.MINUTES)` reads a bit better :laughing:

simplified-profiles/src/main/java/org/nypl/simplified/profiles/ProfileDescriptionJSON.kt
```diff
@@ -257,6 +261,9 @@ object ProfileDescriptionJSON {
     val isManualLCPPassphraseEnabled =
       JSONParserUtilities.getBoolean(objectNode, "isManualLCPPassphraseEnabled")
 
+    val isTimeTrackingEnabled =
+      JSONParserUtilities.getBoolean(objectNode, "isTimeTrackingEnabled")
+
```
io7m, line 266: I think this should be `getBooleanDefault`. I think the line above is actually also a bug... The reason being that the format version being used here is `20191201` and those flags were definitely created _after_ any files were created using that format version. The format version is supposed to be incremented every time there's a backwards-incompatible change. The `getBoolean` method tries to read a boolean-typed field unconditionally and raises an exception if there isn't one...

simplified-books-time-tracking/README.md
nunommts: I was just trying to keep things interesting and mysterious :D

simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingService.kt
nunommts: Yeah, I know. I didn't use the `CompositeDisposable` because I use one of the disposables more than once (to compare, set as null, reinitialize) so I didn't find the need of adding another object just to dispose them. However, since the `remoteStorageDisposable` variable is just used to be disposed, I can add the `CompositeDisposable` to handle it and drop this one.

PR #202, opened 2023-08-15
PP-328: Audio book time tracking is on!

io7m:
**What's this do?**

~This sets the default for audio book time tracking to be on instead of off. Note that, for installations that have been used during development of the time tracking feature, there might already be a setting of "off" saved to the device, so this will require manually toggling the setting back on again. We have no way to know if the setting is "off" because the user turned it off, or if it's "off" because that was the default and the user never even saw the debug settings screen. This shouldn't affect anyone upgrading from a production release, or any new installs.~

This turns on time tracking unconditionally. Any preferences to turn it off have been removed, along with the settings in the debug menu.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-328

Check that the setting is on, on a fresh install or an upgrade from the Play Store release.

None.

Nope.

Nope.

@io7m tried it.

### DIFF
--- a/simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingService.kt
+++ b/simplified-books-time-tracking/src/main/java/org/nypl/simplified/books/time/tracking/TimeTrackingService.kt
@@ -66,10 +66,6 @@ class TimeTrackingService(
     libraryId: String,
     timeTrackingUri: URI?
   ) {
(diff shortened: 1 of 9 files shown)


ray-lee (approved):
Looks good.

ray-lee:
Oh, looks like a test needs to be updated.

io7m:
Ugh, I forgot Gradle doesn't build those unless you specifically ask for them... :roll_eyes:

PR #203, opened 2023-08-16
PP-186: Refactor bookmark service

io7m:
**What's this do?**
This creates a new "create bookmark" operation in the bookmark service that will first create a local bookmark, and then try to create an equivalent remote bookmark. The operation can be configured to ignore remote failures. This refactors the existing operations to return either non-nullable Bookmarks, or fail the resulting Future values with exceptions.

This also refactors the "delete bookmark" operation similarly. The deletion occurs in reverse order; a remote bookmark is deleted first, and then the corresponding local bookmark is deleted if the remote bookmark deletion succeeds. This is also, currently, configured to ignore remote failures at the current call sites.

These changes allow for having the bookmark service do the heavy lifting, and simplifies the various viewers that use it.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-186

Try creating and deleting bookmarks in the audio book player and epub viewer, both with and without annotation syncing enabled.

None.

It was updated in the previous bookmark PR.

Only comments.

@io7m tried various audio books and epubs.


simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -971,22 +971,18 @@ class AudioBookPlayerActivity :
...
-        bookmark = bookmark
-      )
-
-      if (wasDeleted) {
-        currentBookmarks.remove(bookmark)
+    this.bookmarkService.bookmarkDelete(
+      accountID = this.parameters.accountID,
+      bookmark = bookmark,
+      ignoreRemoteFailures = true
+    ).map { wasDeleted ->
```
nunommts, line 978: This `wasDeleted` variable doesn't need to exist now the method is a Unit and doesn't return a boolean, right?

simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/Reader2Activity.kt
```diff
@@ -463,11 +465,13 @@ class Reader2Activity : AppCompatActivity(R.layout.reader2) {
...
         bookmarkService.bookmarkDelete(
           accountID = account.id,
-          bookmark = bookmark
-        ).onAnyError { false }
-          .map { wasDeleted ->
-            event.onDeleteOperationFinished(wasDeleted)
-          }
+          bookmark = bookmark,
+          ignoreRemoteFailures = true
+        ).map { wasDeleted ->
```
nunommts, line 470: Same here!

nunommts (changes requested):
This looks good! I just have two minor comments/suggestions and a more "overall" question:

1 - Is there any reason to have the `ignoreRemoteFailures`variable since it's always true?

io7m:
> This looks good! I just have two minor comments/suggestions and a more "overall" question:
> 
> 1 - Is there any reason to have the `ignoreRemoteFailures`variable since it's always true?

I'm not sure there's a good answer. I was trying to avoid changing the policy too much, but combining the local and remote creation of bookmarks into a single service method meant that it would no longer be possible to distinguish between either one failing; this would mean that in the pretty common case of a remote bookmark failing, no local bookmark would appear to have been created. I thought it might have been a step too far to have the code unconditionally ignore any and every remote failure, so I left it parameterized.

My guess is that we'll be revisiting all of this soon, given that the default is changing and we have no behaviour spec with regards to how the apps should behave when things fail... :sweat:

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
io7m: Oops!

nunommts:
@io7m ok, got it. Since it's fairly easy to remove in the future I don't see any harm on keeping it as it is right now, specially if you're planning to revisit this soon.

nunommts (approved):

PR #206, opened 2023-08-24
Feature/upgrade kotlin version

nunommts:
**What's this do?**
This PR updates the Kotlin version to the 1.9.0 release

There's a submodule (for now it's just one but the goal is to update all of them) that's already compiled with the 1.9.0 version which was causing the `android-core` to fail to compile

The app should compile

None

Yes

No

@nunommts has compiled the app


simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookContentProtections.kt
```diff
@@ -37,6 +37,9 @@ object BookContentProtections {
             provider.isManualPassphraseEnabled = isManualPassphraseEnabled
             provider.onLcpDialogDismissed = onLCPDialogDismissed
           }
+          else -> {
+            // do nothing
```
io7m, line 41: It's best to avoid `else` in `when` expressions. Why? It hides a future bug: At a later date, someone adds another case to the sealed class, and then the compiler can't tell you about `when` expressions that might need updating because you used `else`. If you enumerate the cases explicitly, the compiler can tell you "You've added a new case, this `when` expression is no longer exhaustive!".

simplified-main/src/main/java/org/nypl/simplified/main/MainFragment.kt
```diff
@@ -262,6 +262,9 @@ class MainFragment : Fragment(R.layout.main_tabbed_host) {
       is BookStatus.DownloadWaitingForExternalAuthentication -> {
         this.openBookDownloadLogin(status.id, status.downloadURI)
       }
+      else -> {
```
io7m, line 265: As above.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedViewModel.kt
```diff
@@ -230,11 +236,16 @@ class CatalogFeedViewModel(
...
       when (event.statusNow) {
-        is BookStatus.Held, is BookStatus.Loaned, is BookStatus.Revoked -> {
+        is BookStatus.Held,
+        is BookStatus.Loaned,
+        is BookStatus.Revoked -> {
           if (this.state.arguments.isLocallyGenerated) {
             this.reloadFeed()
           }
         }
+        else -> {
```
io7m, line 246: And here.

io7m (changes requested):
Just needs a couple of small things.

simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookContentProtections.kt
nunommts: You're absolutely right!

io7m (approved):
Seems good!

PR #207, opened 2023-08-28
Fixed crash when opening an audiobook with a null manifest

nunommts:
**What's this do?**
This PR adds a fix for a [recurring crash](https://console.firebase.google.com/project/the-palace-project/crashlytics/app/android:org.thepalaceproject.palace/issues/4ad2d33871ca99f04955282e20c54b32) when trying to open a an audiobook with a null manifest.

There's been a peak of crashes, as reported [here](https://ebce-lyrasis.atlassian.net/browse/PP-382) related to this issue.

I'm not sure in which book(s) this's been happening with, so I don't know exactly how to replicate this issue, but trying to open a significant number of audiobooks should not result in a crash.

None

Yes

No

Not exactly

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -298,9 +298,10 @@
         <c:change date="2023-08-12T00:00:00+00:00" summary="Fixed crash when time tracking is not enabled."/>
       </c:changes>
     </c:release>
-    <c:release date="2023-08-24T16:09:29+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.6.0">
+    <c:release date="2023-08-28T15:28:36+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.6.0">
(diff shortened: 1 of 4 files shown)


io7m:
I'm kind of conflicted about this one. I don't like that a previously non-nullable parameter is now nullable. If the incoming value is non-null, then we already know something is wrong. It's not great that we then pass along this known-wrong value.

io7m:
... Although it's not great that they won't get an error in that case.

:stuck_out_tongue_closed_eyes:

io7m:
... Maybe we merge this as-is, but put a ticket in to refactor this API. It needs to be able to properly signal an error to the caller.

io7m (approved):
This is a disapproving approval.

io7m:
https://ebce-lyrasis.atlassian.net/browse/PP-388

PR #212, opened 2023-09-11
Added support to push notifications and FCM token retrieval.

nunommts:
**What's this do?**
This PR adds support to push notifications as well as some logic to retrieve a FCM token and check if it's already saved on an account's and update the account if it's not. This PR also replaces/deletes the old notifications logic as it wasn't used any longer.

There's a [feature request](https://ebce-lyrasis.atlassian.net/browse/PP-244) to include the push notifications in the Android app.

Log in two different accounts but in the same library

None

Yes

No

Tested by @nunommts


io7m (approved):
I've failed to find anything wrong with this. :+1:

Krsudwa:
.

PR #213, opened 2023-09-13
PP-407: MDC and Crashlytics

io7m:
**What's this do?**
This inserts calls to the `SLFJ4` `MDC` API to append extra metadata to log events. The metadata is then added to any events we log to Crashlytics.

This is likely to be an ongoing thing as we find more metadata that we want to add, and more places where we want it!

https://ebce-lyrasis.atlassian.net/browse/PP-407

It's difficult to test. We need to keep an eye on Crashlytics and see how the new metadata appears in the dashboard.

None.

N/A; internal only.

Only comments.

@io7m Ran the app and tests


io7m:
In the process, I also found and removed some unused test suite code.

simplified-tests/src/test/java/org/nypl/simplified/tests/books/controller/BookRevokeTaskAdobeDRMTest.kt
```diff
@@ -214,9 +216,9 @@ class BookRevokeTaskAdobeDRMTest {
   @Test
   fun testRevokeDRMNonReturnable() {
     val account =
-      Mockito.mock(AccountType::class.java)
+      fakeAccount()
     val accountProvider =
-      Mockito.mock(AccountProviderType::class.java)
+      fakeAccountProvider()
```
nunommts, line 221: Maybe we could declare these two as global variables and initialize them on the @before method instead of declaring and initializing them in all the tests.

simplified-tests/src/test/java/org/nypl/simplified/tests/books/controller/BookRevokeTaskTest.kt
```diff
@@ -196,9 +198,9 @@ class BookRevokeTaskTest {
   @Test
   fun testRevokeNothing() {
     val account =
-      Mockito.mock(AccountType::class.java)
+      fakeAccount()
     val accountProvider =
-      Mockito.mock(AccountProviderType::class.java)
+      fakeAccountProvider()
```
nunommts, line 203: Same thing here

simplified-tests/src/test/java/org/nypl/simplified/tests/books/profiles/ProfileAccountLoginTaskContract.kt
```diff
@@ -183,7 +185,7 @@ abstract class ProfileAccountLoginTaskContract {
       )
 
     val provider =
-      Mockito.mock(AccountProviderType::class.java)
+      fakeAccountProvider()
```
nunommts, line 188: And here

simplified-tests/src/test/java/org/nypl/simplified/tests/books/profiles/ProfileAccountLogoutTaskContract.kt
```diff
@@ -157,7 +158,7 @@ abstract class ProfileAccountLogoutTaskContract {
   @Test
   fun testLogoutNotRequired() {
     val provider =
-      Mockito.mock(AccountProviderType::class.java)
+      fakeAccountProvider()
```
nunommts, line 161: And here

nunommts (approved):
This looks good and I'm approving it. I just left a few suggestions (all related to the same thing) on some of the test classes

io7m:
I agree, but I think I'm going to leave off addressing it in this PR as we have a _lot_ of problems like this in the test suite and anything I did here would only be a half measure.

PR #218, opened 2023-09-18
PP-406: Initialize services eagerly

io7m:
**What's this do?**
This changes the audio book player activity to initialize services eagerly so that they're available to the error handling methods should the activity be called without usable parameters.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-406

Unfortunately, we don't know why the audio book player activity was being called without usable parameters, because we don't yet know why some manifests are resulting in unexpected null values.

None.

Yes.

N/A

@io7m Verified that the audio book player still works.


nunommts (approved):
Looks good!

PR #221, opened 2023-09-18
Ensure the login form is cleared after logout

io7m:
**What's this do?**
This ensures the login form is cleared when logging out.
Additionally, some unused code is removed.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-451

Try logging in and logging out. The username/password fields should be empty afterwards.

None.

No, the problem wasn't present in the last release.

N/A

@io7m Tried it on a couple of accounts


nunommts (approved):
Looks good!

PR #225, opened 2023-09-20
PP-443: Restrict cover widths

io7m:
**What's this do?**
This sets maximum widths for covers so that very large covers can no longer break the UI.

Before:

Screenshot_20230911_104338

Screenshot_20230911_104354

After:

Screenshot_20230920_142139

Screenshot_20230920_142525

https://ebce-lyrasis.atlassian.net/browse/PP-443

Try Minotaur. The book is on the front page now.

None.

Yes.

N/A

@io7m Took the screenshots above!

### DIFF
--- a/simplified-ui-catalog/src/main/res/layout/feed_lane_item.xml
+++ b/simplified-ui-catalog/src/main/res/layout/feed_lane_item.xml
@@ -1,27 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.cardview.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="wrap_content"
-    android:layout_height="wrap_content"
-    android:clipChildren="false"
-    android:clipToPadding="false"
-    android:orientation="vertical">
+    android:layout_height="@dimen/cover_thumbnail_height"
+    app:cardCornerRadius="@dimen/cover_corner_radius"
+    app:cardElevation="2dp">
 
-    <androidx.cardview.widget.CardView
+    <ImageView
+        android:id="@+id/coverImage"
         android:layout_width="wrap_content"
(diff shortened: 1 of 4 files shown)


simplified-ui-catalog/src/main/res/layout/feed_lane_item.xml
```diff
@@ -1,25 +1,27 @@
...
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
-  xmlns:app="http://schemas.android.com/apk/res-auto"
-  android:layout_width="wrap_content"
-  android:layout_height="wrap_content"
-  android:clipChildren="false"
-  android:clipToPadding="false"
-  android:orientation="vertical">
 
-  <androidx.cardview.widget.CardView
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
nunommts, line 3: Since there's just a CardView, do we need a parent view? If so, maybe we should use a FrameLayout instead then

nunommts (changes requested):
Just a comment on a minor issue

nunommts (changes requested):
Just a comment on a minor issue

simplified-ui-catalog/src/main/res/layout/feed_lane_item.xml
io7m: I freely admit I have no idea. I've not looked at any of the UI code for years, and even then, I wasn't the person who wrote it.

What arrangement of layouts were you thinking of?

```
<FrameLayout>
  <CardView>
    <ImageView/>
  </CardView>
</FrameLayout>
```

Like that?

io7m: Removing the parent view completely seems to work fine!

nunommts: Cool! But for the record, I was thinking about this (like you said), if we were not to remove the parent view:

```
<FrameLayout>
  <CardView>
    <ImageView/>
  </CardView>
</FrameLayout>
```

nunommts (approved):

PR #227, opened 2023-09-21
PP-307: Resume account views correctly

io7m:
**What's this do?**
The current code was failing to set usernames/passwords for token auth,
and also failing to reconfigure the UI in the `onStart()` method. This
would cause inconsistent visuals when resuming the app after having
locked the screen.

https://ebce-lyrasis.atlassian.net/browse/PP-307

Try locking the device when on the account details screen, and then unlocking it. See the video in PP-307 for an example of what _shouldn't_ happen. :slightly_smiling_face: 

None.

Yes.

N/A

@io7m Tried lots of locking and unlocking.


nunommts (approved):
Looks good!

PR #228, opened 2023-09-21
Feature/add support to epub text searching

nunommts:
**What's this do?**
This PR adds support to the EPUB text searching by handling the logic of showing/hiding the SearchFragment

There's a [feature request](https://ebce-lyrasis.atlassian.net/browse/PP-426) to include the EPUB text searching in the Palace app.

Open the Palace app
Select any EPUB
Click on the search icon
Type any query
Click on one of the results
Confirm you're redirected to that part of the book

[This PR](https://github.com/ThePalaceProject/android-r2/pull/32) also needs to be merged

Yes

No

Tested by @nunommts


simplified-viewer-epub-readium2/src/main/java/org/librarysimplified/viewer/epub/readium2/Reader2Activity.kt
```diff
@@ -189,29 +193,25 @@ class Reader2Activity : AppCompatActivity(R.layout.reader2) {
...
-        .commit()
-    } else {
-      this.readerFragment =
-        this.supportFragmentManager.findFragmentByTag(READER_FRAGMENT_TAG) as SR2ReaderFragment
-      this.tocFragment =
-        this.supportFragmentManager.findFragmentByTag(TOC_FRAGMENT_TAG) as SR2TOCFragment
-    }
+    this.supportFragmentManager.beginTransaction()
+      .add(R.id.reader2FragmentHost, this.readerFragment)
+      .commit()
```
io7m, line 214: We aren't reusing fragment instances anymore?

Not that I mind, but ... does this have lifecycle repercussions?

nunommts: We are. This adds the reader fragment and then all the operations on fragments are just showing/hiding the current instances (or add one if it's not added already) so there's no recreation of fragments. In the end, when we leave the screen all the instances are destroyed alongside with the activity

io7m (approved):
:rocket:

PR #232, opened 2023-09-26
Updated OPDS facet parsing by ignoring the content of facetGroupType

nunommts:
**What's this do?**
This PR updates the OPDS facet parsing logic by ignoring the content of the facetGroupType.

There's a [ticket with a bug](https://ebce-lyrasis.atlassian.net/browse/PP-476) saying the catalog is no longer displaying the tabs. The iOS app is still displaying them because they ignore the content of the attribute `facetGroupType` as long as the key is there.

Open the Palace app
Confirm there are tabs in the top of the catalog

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -298,7 +298,7 @@
         <c:change date="2023-08-12T00:00:00+00:00" summary="Fixed crash when time tracking is not enabled."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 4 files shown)


io7m (changes requested):
Can you add a unit test for this one?

Given that there's no spec, it would be nice to have a test that says what we're doing so we know when it changes next.

io7m:
Actually, it looks like a test failed.

io7m (approved):
Seems good!

PR #241, opened 2023-10-16
PP-389: New material theme

io7m:
**What's this do?**
This makes all of the various changes to use the new Palace material theme. The old Simplified theme code and neutral theme are completely removed.

https://ebce-lyrasis.atlassian.net/browse/PP-389

This is a tough one because the changes are all visual. Problems are likely to be of the type "I can't see a UI element that I'm supposed to be see" or "Something is misaligned".

None really, but this will need a lot of informal testing. I'd rather it went out in 1.7.0 than 1.6.1.

Not yet, because we've yet to push 1.6.1, and this will probably be 1.7.0.

N/A

@io7m Ran the app and the various viewers a lot.


io7m:
The build is failing because we need to deploy snapshots of the audiobooks and r2 components to Central. We'll need to coordinate this, because we don't want those changes going out in 1.6.1...

io7m:
I've just realized I missed out the tabbed button changes. I'll add these today.

simplified-viewer-audiobook/src/main/java/org/librarysimplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -848,7 +848,7 @@ class AudioBookPlayerActivity :
      */
 
     this.uiThread.runOnUIThread {
-      val fragment = PlayerSleepTimerFragment.newInstance()
+      val fragment = PlayerSleepTimerFragment.newInstance(PlayerFragmentParameters())
```
nunommts, line 851: This shouldn't be here because when the audiobook submodule versions are correctly implemented and imported, we'll have an error

nunommts (changes requested):
This looks really good! I have just one small comment on the PlayerTimer argument change

io7m:
Done.

PR #242, opened 2023-10-17
Adjusted feed tabs navigation to keep the "Catalog" on the screen

nunommts:
**What's this do?**
This PR changes how the feed handles a new facet/tab click. Instead of creating and adding a new fragment for the feed, it just updates the UI and reloads the feed with some new arguments.

There's a [feature request](https://ebce-lyrasis.atlassian.net/browse/PP-577) to make this similar as it is on the iOS app

Open the Palace app
Select "LYRASIS Reads"
On the Catalog tab, press "eBooks"
Confirm the title remains as "Catalog" and there's the Palace logo instead of a back arrow
Press "AudioBooks"
Confirm the title remains as "Catalog" and there's the Palace logo instead of a back arrow

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -332,9 +332,10 @@
         <c:change date="2023-10-07T00:00:00+00:00" summary="Fixed Audiobook UI freezing after pressing 'play'."/>
       </c:changes>
     </c:release>
-    <c:release date="2023-10-16T11:36:56+00:00" is-open="true" ticket-system="org.nypl.jira" version="1.6.1">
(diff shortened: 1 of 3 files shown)


io7m:
I'm ... not clear on how this fixes the problem.

nunommts:
@io7m Instead of calling the `this.listener.post(CatalogFeedEvent.OpenFeed(feedArguments))` and consequently adding a new fragment to the stack, the ViewModel will just update the current feed's arguments and run a new search with the updated arguments. Then, the UI will catch the loading state and show the loading progress and update the selected facet. This way we keep things on the same Fragment and the "Catalog" and the Palace logo will always be there!

PR #245, opened 2023-10-24
Enabled push notifications by default

nunommts:
**What's this do?**
This PR enables the push notifications by default and removes the option from the debug settings.

There's a [feature request](https://ebce-lyrasis.atlassian.net/browse/PP-571) to enable the push notifications by default now that they're tested

Log in two different accounts but in the same library

None

Yes

No

Tested by @nunommts


io7m (approved):
:fire:

PR #246, opened 2023-10-25
Fixed crash on Reader screen due to more than one Fragment instance

nunommts:
**What's this do?**
This PR fixes a recurring crash on the ReaderFragment when more than one fragment tries to listen to a ViewModel observable that can only be subscribed once.

There's a [bug report](https://ebce-lyrasis.atlassian.net/browse/PP-633) mentioning there's been a significant number of crashes because of this

First you need to enable the Developer Options of your mobile phone (please Google how to do it as it's different from device to device)
Open the Developer Options
Enable the option "Don't keep activities"
Open the Palace app
Open any Ebook
Minimize the app
Reopen the app
Confirm the app didn't crash

None

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -337,7 +337,7 @@
         <c:change date="2023-10-16T00:00:00+00:00" summary="Fixed crash when creating a library card."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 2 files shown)


io7m (approved):
I hate fragments. :+1:

PR #247, opened 2023-10-27
PP-*: Use material dialogs everywhere

io7m:
**What's this do?**
This switches to using the `MaterialAlertDialog` everywhere.
This should address a number of theme-related issues.

We just made this change in the other components. These dialogs are causing the same issue in the core app too, but there's no open ticket specifically about that.

Related: https://ebce-lyrasis.atlassian.net/browse/PP-651
Related: https://ebce-lyrasis.atlassian.net/browse/PP-650

Check that A1QA don't file new tickets about dialogs. :slightly_smiling_face: 

None.

N/A

N/A

@io7m Opened a few of the dialogs.


nunommts (approved):
Update everything! \o/

PR #253, opened 2023-11-07
PP-642: No previews for loans

io7m:
**What's this do?**
This adjusts the code so that the "Read Sample" button is no longer present for books that are already loaned.

https://ebce-lyrasis.atlassian.net/browse/PP-642

Most of the Palace Marketplace books on Minotaur have samples. Check that you don't see a Read Sample button in the book detail page when you've borrowed a book.

None.

Yes.

N/A

@io7m Tried it on the Palace Marketplace.

### DIFF
--- a/simplified-ui-catalog/src/main/java/org/librarysimplified/ui/catalog/CatalogBookDetailFragment.kt
+++ b/simplified-ui-catalog/src/main/java/org/librarysimplified/ui/catalog/CatalogBookDetailFragment.kt
@@ -453,7 +453,7 @@ class CatalogBookDetailFragment : Fragment(R.layout.book_detail) {
         this.onBookStatusHeld(status, bookPreviewStatus)
       }
       is BookStatus.Loaned -> {
-        this.onBookStatusLoaned(status, book.book, bookPreviewStatus)
+        this.onBookStatusLoaned(status, book.book)
       }
       is BookStatus.Holdable -> {
         this.onBookStatusHoldable(status, bookPreviewStatus)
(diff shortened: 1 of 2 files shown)


simplified-ui-catalog/src/main/java/org/librarysimplified/ui/catalog/CatalogBookDetailFragment.kt
```diff
@@ -912,10 +895,6 @@ class CatalogBookDetailFragment : Fragment(R.layout.book_detail) {
           }
         )
       )
-    } else if (!createPreviewButton) {
-      // add spaces on both sides if there aren't any other buttons
-      this.buttons.addView(this.buttonCreator.createButtonSizedSpace(), 0)
-      this.buttons.addView(this.buttonCreator.createButtonSizedSpace())
```
nunommts, line 918: I think that removing these two buttons is causing [this](https://github.com/ThePalaceProject/android-core/assets/79104027/83f9d8d5-d241-41d5-af69-eaabe0f94836). If you just remove the condition of the `else if` and keep the rest, it should be fixed. Also, I think you can remove the `bookPreviewStatus` as argument of this method since it's no longer used

nunommts (changes requested):

simplified-ui-catalog/src/main/java/org/librarysimplified/ui/catalog/CatalogBookDetailFragment.kt
io7m: Oops!

nunommts (approved):

PR #254, opened 2023-11-07
PP-693 (partial): Switch to a new ULID library

io7m:
**What's this do?**
This is a partial fix for PP-693. We were using a ULID library that can't work on old devices due to them missing the standard `java.time` package.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-693

Unfortunately, the player still appears to be broken (and I'm debugging right now). This at least makes it slightly less broken.

None.

Not until the problem is actually fixed.

N/A

@io7m Verified one of the problems went away.


nunommts (approved):
Farewell, my ULID library discovery! \o/

PR #255, opened 2023-11-08
Started using device registration URI from patron authentication document

nunommts:
**What's this do?**
This PR adds logic to use the device registration URI, if any, on the patron authentication document instead of having it hardcoded. It also changes the notification http calls to be done on a background thread.

There's a [feature request](https://ebce-lyrasis.atlassian.net/browse/PP-654) to implement this now the backend supports it and there's a [bug report](https://ebce-lyrasis.atlassian.net/browse/PP-705) saying the notification http calls are being called in the main thread.

Logout and login (or just log in if you were not authenticated) from any library (but it has to be the same) on both devices

None

Yes

No

Tested by @nunommts

### DIFF
--- a/simplified-notifications/src/main/java/NotificationTokenHTTPCalls.kt
+++ b/simplified-notifications/src/main/java/NotificationTokenHTTPCalls.kt
@@ -35,21 +35,24 @@ class NotificationTokenHTTPCalls(
   override fun registerFCMTokenForProfileAccount(
     account: AccountType
   ) {
+    val credentials = account.loginState.credentials
+    val originalUrl = credentials?.deviceRegistrationURI
+
+    if (originalUrl == null) {
+      logger.debug("Account {} doesn't have a device registration uri", account.id)
+      return
+    }
+
     firebaseInstance.token
       .addOnSuccessListener { token ->
         logger.debug("Success fetching FCM Token: {}", token)
-        val originalUrl = buildString {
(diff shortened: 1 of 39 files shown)


io7m:
Is this actually still relevant? The platform has been updated multiple times since this was submitted.

nunommts:
@io7m hmm, you're right. If there's a more recent version of the platform then this is not needed. However, I'll use and adapt this for my next PR :D

nunommts:
@io7m it's ready for review 😄

simplified-notifications/src/main/java/NotificationTokenHTTPCalls.kt
```diff
@@ -35,61 +37,66 @@ class NotificationTokenHTTPCalls(
...
-              )
-              setFCMTokenForAccount(
-                account = account,
-                token = token,
-                url = originalUrl
-              )
-            } else {
+            is LSHTTPResponseStatus.Responded.Error -> {
+              if (status.properties.status == HttpURLConnection.HTTP_NOT_FOUND) {
+                logger.error(
```
io7m, line 74: I don't think this should be at `error` severity. It'll cause a non-fatal event to be logged to Crashlytics. This is probably just `debug`.

io7m (changes requested):
Seems good, just one small adjustment needed.

io7m (approved):
:rocket:

PR #258, opened 2023-11-10
PP-714: Make BottomNavigator bug non-fatal

io7m:
**What's this do?**
Don't attempt to pop the stack unless there's something to pop. 
Catch any NullPointerExceptions that the library might raise.

Affects: https://ebce-lyrasis.atlassian.net/browse/PP-714

Nothing new. Just see if the crashes stop after `1.8.0`...

None.

Yes.

N/A

@io7m Used the catalog for a while. I've not seen the original crash, so I just wanted to check that I didn't break the catalog.

### DIFF
--- a/simplified-ui-navigation-tabs/src/main/java/org/librarysimplified/ui/navigation/tabs/TabbedNavigator.kt
+++ b/simplified-ui-navigation-tabs/src/main/java/org/librarysimplified/ui/navigation/tabs/TabbedNavigator.kt
@@ -70,7 +70,16 @@ class TabbedNavigator private constructor(private val navigator: BottomNavigator
   }
 
   fun popBackStack(): Boolean {
-    return this.navigator.pop()
+    if (this.navigator.currentStackSize() < 1) {
+      return false
+    }
+
+    return try {
+      this.navigator.pop()
+    } catch (e: Exception) {
+      logger.warn("Failed to pop navigator stack: ", e)
+      false
+    }
   }
 
   fun popToRoot(): Boolean {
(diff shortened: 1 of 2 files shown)


simplified-ui-navigation-tabs/src/main/java/org/librarysimplified/ui/navigation/tabs/TabbedNavigator.kt
```diff
@@ -70,7 +70,16 @@ class TabbedNavigator private constructor(private val navigator: BottomNavigator
   }
 
   fun popBackStack(): Boolean {
-    return this.navigator.pop()
+    if (this.navigator.currentStackSize() < 1) {
+      return false
```
nunommts, line 74: Does the `this.navigator.pop()` already handle this? I imagine it doesn't, so it would be helpful to add a log here too so we can be aware if it's possible to call this even when there's nothing to pop. Also, you could add this to the `try-catch`

nunommts (changes requested):
Looks good. Just a minor suggestion, though.

simplified-ui-navigation-tabs/src/main/java/org/librarysimplified/ui/navigation/tabs/TabbedNavigator.kt
io7m: This was just pure paranoia. I think it crashes instead due to an internal bug.

io7m: Adjusted!

nunommts (approved):
Let's go!

PR #261, opened 2023-11-17
Updated Readium version

nunommts:
**What's this do?**
This PR updates the app to support the latest version of Readium

There's a [feature request](https://ebce-lyrasis.atlassian.net/browse/PP-704) to update the Readium version

Open the Palace app
Open any Ebook
Confirm everything is working
Return and open any PDF
Confirm everything is working

[This PR](https://github.com/ThePalaceProject/android-r2/pull/55) also needs to be merged

Yes

No

Tested by @nunommts

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -370,7 +370,7 @@
         <c:change date="2022-04-29T00:00:00+00:00" summary="Fixed &quot;Unable to initialize audio engine&quot; error playing some audio books."/>
       </c:changes>
     </c:release>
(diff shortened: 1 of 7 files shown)


io7m:
This looks fine to me, but I must admit that I have no idea what's going on with our PDF support. How is Readium involved? I thought we used some other third party library.

nunommts:
@io7m We're using their `PdfParser` file. One of the main changes between the previous Readium version and this one is that the `PdfParser` now doesn't accept a constructor with no PDFFactory argument that would use their "default" document factory, so I had to import the code they're using on it.

PR #266, opened 2023-11-21
PP-747: Make bookmark syncing a purely client-side option

io7m:
**What's this do?**
This removes all code relating to the server-side bookmarks setting, and moves to having the on/off switch be purely local state.

Fix: https://ebce-lyrasis.atlassian.net/browse/PP-747

Check that no bookmarks are logged if the switch is off. Check that they _are_ if the switch is on.
Note that this will currently "fail" on real servers (sending bookmarks will probably receive a 403). This depends on a server-side change to work fully.

Yes, we can't release this until the server-side work is done.

Yes

Yes

@io7m Tried it!


nunommts (approved):
Oh, the sweet sound of a lot of code deletion... <3

io7m:
Server side was just deployed.

PR #269, opened 2023-11-27
PP-756, PP-757: Adjust account deletion UI

io7m:
**What's this do?**
* Remove the account deletion popup menu in favour of a plain dialog popup.
* Fix the missing error message when trying to delete the last account.
* Use "Remove" instead of "Delete" for consistency.

Fix: https://ebce-lyrasis.atlassian.net/browse/PP-756

These are covered by A1QA, apparently!

None.

Yes.

N/A

@io7m Tried it.

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -407,7 +407,7 @@
         </c:change>
       </c:changes>
     </c:release>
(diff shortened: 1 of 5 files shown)


nunommts (approved):
Nice!

PR #304, opened 2026-04-14
Replace softprops/action-gh-release with direct GH api create release…

dbernstein:
… call

We're trying to get the jira-release-sync workflow to trigger on a "publish" event.

Should be tested during next release.
No.

Not yet: but it was taken from ios-core where it has run successfully.

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -17,18 +17,15 @@ jobs:
         run: .ci/ci-verify-release-branch.sh
       - name: Finish release
         run: .ci/ci-release-finish.sh
-      - name: Create GitHub release
-        uses: softprops/action-gh-release@v2
+      - name: Create or update release
         env:
+          GITHUB_TOKEN: ${{ GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/android-release.yml
```diff
@@ -17,18 +17,15 @@ jobs:
         run: .ci/ci-verify-release-branch.sh
       - name: Finish release
         run: .ci/ci-release-finish.sh
-      - name: Create GitHub release
-        uses: softprops/action-gh-release@v2
+      - name: Create or update release
         env:
+          GITHUB_TOKEN: ${{ GITHUB_TOKEN }}
```
io7m: I think that should be `secrets.CI_GITHUB_ACCESS_TOKEN`.

io7m (changes requested):
GitHub seems unhappy with this. Comes up with an error in the YAML viewer.

.github/workflows/android-release.yml
dbernstein: yeah - I wasn't sure since ios-core wasn't using it. But it's probably better to go with the ci token.

io7m (approved):
It seems happy with that. 🚀

